PR #363, opened 2026-02-17
Add ProcessTaskExecutor

### LINKED ISSUE #82: New OWEwoksWidget to run tasks in a process
***In GitLab by @loichuder on Jan 23, 2026, 15:09 GMT+1:***

`OWEwoksWidgetOneThread` is a great abstraction allowing to run the task in a thread. This gives an easy way of having a responsive GUI while the task is running.

Unfortunately, threads cannot be killed so aborting/canceling the task is tricky.

I would propose to create a new Orange Ewoks widget to run the tasks in a `multiprocess.Process` instead. With a bit a care, processes can be killed to cancel ongoing tasks. This new widget could provide a nice abstraction like `OWEwoksWidgetOneThread` so that ewoksapps devs do not need to deal with implementation details.

Thoughts @woutdenolf @payno ?

*Migrated from GitLab: https://gitlab.esrf.fr/workflow/ewoks/ewoksorange/-/issues/82*

loichuder:
Fix #82

### DIFF
--- a/src/ewoksorange/gui/concurrency/process.py
+++ b/src/ewoksorange/gui/concurrency/process.py
@@ -1,0 +1,36 @@
+import subprocess
+from typing import Optional, Type
+
+from ewokscore.task import Task
+
+from .base import TaskExecutor
+
+
+class ProcessTaskExecutor(TaskExecutor):
+    """Create and execute an Ewoks task in a process."""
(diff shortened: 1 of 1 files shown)


loichuder:
Superseded by #420

PR #381, opened 2026-04-13
Feature/filter widget registry

abmajith:
Screenshot from 2026-04-13 14-57-15
```bash
EWOKS_WIDGET_FILTER="Ewoks Demo" ewoks-canvas
EWOKS_WIDGET_FILTER="None" ewoks-canvas
```

select the required data processing group box (left side of the orange canvas), to limit the number of data processing groups


woutdenolf:
The Ewoks demo is no longer visible by default. So this is no longer needed right?

abmajith:
> The Ewoks demo is no longer visible by default. So this is no longer needed right?

Yes, we dont need this one, its already solved in the release

PR #386, opened 2026-05-19
Draft: canvas: add 'Trigger' action to the link

payno:
Note to reviewer(s): this PR needs to be updated before merged (once the orangecanvas is released - see [associated note](https://github.com/ewoks-kit/ewoksorange/pull/386/changes#r3266518973)).

The rest of the code will not be modified.

### DIFF
--- a/src/ewoksorange/gui/canvas/main.py
+++ b/src/ewoksorange/gui/canvas/main.py
@@ -2,6 +2,8 @@ import logging
 import sys
 from contextlib import contextmanager
 
+from packaging.pylock import _logger
+
 from ...orange_version import ORANGE_VERSION
 
 if ORANGE_VERSION == ORANGE_VERSION.oasys_fork:
@@ -74,8 +76,23 @@ elif ORANGE_VERSION == ORANGE_VERSION.latest_orange:
     from Orange.canvas.__main__ import main as _main
     from orangecanvas.main import arg_parser
 else:
+    import orangecanvas.main
     from orangecanvas.main import arg_parser
-    from orangecanvas.main import main as _main
+    from .mainwindow import CanvasMainWindow
+
+    if not hasattr(CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
+        _logger.info(
+            "Canvas main window does not have an editor widget constructor, skipping adding the 'trigger' action to links"
+        )
+
+    class Main(orangecanvas.main.Main):
+
+        def create_main_window(self) -> CanvasMainWindow:
+            """Create the (initial) main window."""
+            return CanvasMainWindow()
+
+    def _main(argv=None):
+        return Main().run(argv)
 
 
 @contextmanager
--- a/src/ewoksorange/gui/canvas/main.py
+++ b/src/ewoksorange/gui/canvas/main.py
@@ -77,20 +77,25 @@ elif ORANGE_VERSION == ORANGE_VERSION.latest_orange:
     from orangecanvas.main import arg_parser
 else:
     import orangecanvas.main
+    from orangecanvas.application.canvasmain import (
+        CanvasMainWindow as _CanvasMainWindow,
+    )
     from orangecanvas.main import arg_parser
 
-    from .mainwindow import CanvasMainWindow
+    from .schemeedit import SchemeEditWidget as _SchemeEditWidget
 
-    if not hasattr(CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
+    if not hasattr(_CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
         _logger.info(
             "Canvas main window does not have an editor widget constructor, skipping adding the 'trigger' action to links"
         )
 
     class Main(orangecanvas.main.Main):
 
-        def create_main_window(self) -> CanvasMainWindow:
+        def create_main_window(self) -> _CanvasMainWindow:
             """Create the (initial) main window."""
-            return CanvasMainWindow()
+            _CanvasMainWindow.EDITOR_WIDGET_CONSTRUCTOR = _SchemeEditWidget
+            # Redefining the 'EDITOR_WIDGET_CONSTRUCTOR' allow use to add features to the default 'SchemeEditWidget' like adding actions to the link.
+            return _CanvasMainWindow()
 
     def _main(argv=None):
         return Main().run(argv)
--- a/src/ewoksorange/gui/canvas/main.py
+++ b/src/ewoksorange/gui/canvas/main.py
@@ -78,6 +78,7 @@ elif ORANGE_VERSION == ORANGE_VERSION.latest_orange:
 else:
     import orangecanvas.main
     from orangecanvas.main import arg_parser
+
     from .mainwindow import CanvasMainWindow
 
     if not hasattr(CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
--- a/src/ewoksorange/gui/canvas/main.py
+++ b/src/ewoksorange/gui/canvas/main.py
@@ -2,9 +2,9 @@ import logging
 import sys
 from contextlib import contextmanager
 
-from packaging.pylock import _logger
-
 from ...orange_version import ORANGE_VERSION
+
+_logger = logging.getLogger(__file__)
 
 if ORANGE_VERSION == ORANGE_VERSION.oasys_fork:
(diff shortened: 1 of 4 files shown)


src/ewoksorange/gui/canvas/main.py
```diff
@@ -74,8 +76,23 @@ def arg_parser():
...
 else:
+    import orangecanvas.main
     from orangecanvas.main import arg_parser
-    from orangecanvas.main import main as _main
+    from .mainwindow import CanvasMainWindow
+
+    if not hasattr(CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
+        _logger.info(
+            "Canvas main window does not have an editor widget constructor, skipping adding the 'trigger' action to links"
+        )
```
payno: This should be replaced by checking the orangecanvas version once https://github.com/biolab/orange-canvas-core/pull/349 merged / integrated.
En then merge. There is no urgency.

src/ewoksorange/gui/canvas/main.py
```diff
@@ -74,8 +76,23 @@ def arg_parser():
...
+
+    class Main(orangecanvas.main.Main):
+
+        def create_main_window(self) -> CanvasMainWindow:
+            """Create the (initial) main window."""
+            return CanvasMainWindow()
+
+    def _main(argv=None):
+        return Main().run(argv)
```
payno: Maybe this part can be improved to avoid defining the Main in the `else` section. Looks a bit noisy. But simpler to keep handing the `oasys` fork.

src/ewoksorange/gui/canvas/schemeedit.py
```diff
@@ -0,0 +1,32 @@
...
+            triggered=self.__linkTrigger,
+            toolTip=self.tr("Trigger link."),
+        )
+
+        linkMenu = self.linkMenu()
+        linkMenu.addSeparator()
+        linkMenu.addAction(self.__linkTriggerAction)
+
+    def __linkTrigger(self):
+        link = self._SchemeEditWidget__contextMenuTarget
```
payno: At the moment I don't see how to retrieve the link from the public API.

src/ewoksorange/gui/canvas/mainwindow.py
```diff
@@ -0,0 +1,7 @@
+from orangecanvas.application.canvasmain import CanvasMainWindow as _CanvasMainWindow
+from .schemeedit import SchemeEditWidget
+
+
+class CanvasMainWindow(_CanvasMainWindow):
```
payno: The idea is too use the `EDITOR_WIDGET_CONSTRUCTOR` that will be used now by orangecanvas.
For older version it will only be ignored so I don't think it needed to add a constrain on the orangecanvas version as this feature is a 'nice to have' but not mandatory. Can be discussed.

src/ewoksorange/gui/canvas/schemeedit.py
```diff
@@ -0,0 +1,41 @@
+from AnyQt.QtWidgets import QAction, QMenu
+from orangecanvas.document.schemeedit import SchemeEditWidget as _SchemeEditWidget
+from orangecanvas.scheme import SchemeLink
+
+
+class SchemeEditWidget(_SchemeEditWidget):
```
payno, line 11: This class will be instantiate only for version of orangecanvas to come.

src/ewoksorange/gui/canvas/main.py
```diff
@@ -2,6 +2,8 @@
 import sys
 from contextlib import contextmanager
 
+from packaging.pylock import _logger
```
woutdenolf: That's unusual. Also we import a private object.

src/ewoksorange/gui/canvas/mainwindow.py
woutdenolf: Sure, no need to check versions here.

But why do we need to derive a class and not simply set the class attribute `orangecanvas.application.canvasmain.CanvasMainWindow.EDITOR_WIDGET_CONSTRUCTOR = SchemeEditWidget` ?

src/ewoksorange/gui/canvas/schemeedit.py
```diff
@@ -0,0 +1,42 @@
+from AnyQt.QtWidgets import QAction
+from AnyQt.QtWidgets import QMenu
+from orangecanvas.document.schemeedit import SchemeEditWidget as _SchemeEditWidget
```
woutdenolf, line 5: For reference, this import also works for the current orange version.

src/ewoksorange/gui/canvas/schemeedit.py
```diff
@@ -0,0 +1,42 @@
...
+        self.__link = None  # type: Optional[SchemeLink]
+        # Active link in the context menu.
+
+    def __linkTrigger(self):
+        link = self.__link
+        if not link:
+            return
+
+        self.removeLink(link)
+        self.addLink(link)
```
woutdenolf: Will this preserve the data mapping? Is there no cleaner way to cause the triggering?

Perhaps not. I tried following the code to the place where the downstream nodes are triggered (provided their other incoming links are triggered):

- [self.link_added.emit(link)](https://github.com/biolab/orange-canvas-core/blob/3654ee8f4afb2ad404014513547bef16144c4e04/orangecanvas/scheme/scheme.py#L325)
- [self.scheme.link_added.connect(self.add_link)](https://github.com/biolab/orange-canvas-core/blob/3654ee8f4afb2ad404014513547bef16144c4e04/orangecanvas/canvas/scene.py#L210)
- [workflow.link_added.connect(self.__on_link_added)](https://github.com/biolab/orange-canvas-core/blob/3654ee8f4afb2ad404014513547bef16144c4e04/orangecanvas/scheme/signalmanager.py#L402)
- [self._schedule(signals)](https://github.com/biolab/orange-canvas-core/blob/3654ee8f4afb2ad404014513547bef16144c4e04/orangecanvas/scheme/signalmanager.py#L547)

Maybe one of these signals could be emitted?

woutdenolf (changes requested):

src/ewoksorange/gui/canvas/main.py
payno: Oula indeed, looks like my copilot or vscode auto completion is playing games :(. Sorry

src/ewoksorange/gui/canvas/mainwindow.py
payno: Yes we can as well indeed, will be better.
292c408d64fe2ddbc8661f1efaac238b304973dd will take care of it.

src/ewoksorange/gui/canvas/schemeedit.py
payno: Good question(s).

* `Scheme.link_inserted` and `Scheme.link_added` (at least) will be trigger. Because `Scheme.insert_link` will be executed in this case.

* As we are reusing the `SchemeLink` data mapping will be preserved because it contains the mapping under the forms: `source_node`, `source_channel`, `sink_node`, `sink_channel`. Naming is close to ewoks except that `target` is replaced by `sink` in orange.

* On the 'performance' side I am sure something better could be done (reducing internals checks...) but it would force deeper refactoring, comprehension of the code and more complexity. Using it this way I never saw it being very slow. I guess at the moment extra processing time is negligible.

woutdenolf: I was not worried about the performance. As long as there are no side-effect I'm ok with it.

src/ewoksorange/gui/canvas/main.py
```diff
@@ -74,8 +76,29 @@ def arg_parser():
...
+    if not hasattr(_CanvasMainWindow, "EDITOR_WIDGET_CONSTRUCTOR"):
+        _logger.info(
+            "Canvas main window does not have an editor widget constructor, skipping adding the 'trigger' action to links"
+        )
+
+    class Main(orangecanvas.main.Main):
+
+        def create_main_window(self) -> _CanvasMainWindow:
+            """Create the (initial) main window."""
+            _CanvasMainWindow.EDITOR_WIDGET_CONSTRUCTOR = _SchemeEditWidget
```
woutdenolf: Why do we need to derive a class again? Can't we simply set the `EDITOR_WIDGET_CONSTRUCTOR` attribute of the class in `orangecanvas`?

woutdenolf (approved):
Thanks!

woutdenolf:
What is the status of this? It was approved but then made draft again.

PR #390, opened 2026-06-01
ewoksOrange task processing: move to Future.

### LINKED ISSUE #387: OWEwoksWidgetWithTaskStack: provide API to cancel a task
Today `OWEwoksWidgetWithTaskStack` doesn't provide API to cancel a task.

There is two approaches:

* simple: allow to cancel the current task and the clear all task
* advanced but more powerful: allow to cancel a specific task (like using an id)

payno:
# Overview

This PR adds a complete API to handle task cancellation from the different ewoksorange widgets.

The idea is to make to:

* provide a Future (under the form of a `TaskFuture`) to the user when a task is executed.
* This future can be cancelled by the user (either if processing is pending or on-going)
* Adapt existing structures to be able to cancel a specific task and all existing tasks of an executor
* The public API for task cancellation is moved from the different executors to the  ̀Future` object. Only remains for conveniance a `cancel_all_tasks` Moving `cancel_running_task` also ease handling correctly of the `Future.cancel ̀. Because `cancel` must call executor cancellation **and** executor cancellation must make sure `Future.cancel` is called to have a coherent state. This was we only need to handle task cancellation is called ... from the `Future`.

### DIFF
--- a/src/ewoksorange/gui/concurrency/_future.py
+++ b/src/ewoksorange/gui/concurrency/_future.py
@@ -38,7 +38,6 @@ class TaskFuture(_Future):
         if self.done():
             return False
         if self.executor:
-            return self.executor._cancel_future(self)
-        # this should never happen but safer to log this us case
-        _logger.error("executor has been deleted. Unable to cancel.")
-        return False
+            self.executor._cancel_future(self)
+
+        return super().cancel()
--- a/src/ewoksorange/gui/concurrency/_future.py
+++ b/src/ewoksorange/gui/concurrency/_future.py
@@ -1,0 +1,37 @@
+import logging
+from concurrent.futures import Future as _Future
+import weakref
+from .base import TaskExecutionID
+
+_logger = logging.getLogger(__name__)
+
+
+class ExecutorFutureHandler:
+    """Class allowing handling of 'TaskFuture'"""
+
+    def cancel_task(self, task_exec_id: TaskExecutionID) -> None:
+        raise NotImplementedError("Base class")
+
+
+class TaskFuture(_Future):
+    """Implementation of Future for tasks and 'ExecutorFutureHandler'"""
+
+    def __init__(self, task_exec_id: str, executor: ExecutorFutureHandler, **kwargs):
+        super().__init__()
+
+        self._executor = weakref.ref(executor)
+        self.task_kwargs = {}
+        self.task_exec_id = task_exec_id
+
+    @property
+    def executor(self) -> ExecutorFutureHandler | None:
+        return self._executor()
+
+    def cancel(self) -> bool:
+        if self.done():
+            return False
+        if self.executor:
+            return self.executor._cancel_future(self)
+        # this should never happen but safer to log this us case
+        _logger.error("executor has been deleted. Unable to cancel.")
+        return False
--- a/src/ewoksorange/gui/concurrency/_future.py
+++ b/src/ewoksorange/gui/concurrency/_future.py
@@ -1,23 +1,30 @@
 import logging
-from concurrent.futures import Future as _Future
 import weakref
-from .base import TaskExecutionID
+from concurrent.futures import Future as _Future
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from .base import TaskExecutionID
 
 _logger = logging.getLogger(__name__)
 
 
 class ExecutorFutureHandler:
-    """Class allowing handling of 'TaskFuture'"""
+    """Define internal API to cancel a future."""
 
-    def cancel_task(self, task_exec_id: TaskExecutionID) -> None:
+    def _cancel_future(self, future: TaskFuture) -> None:
         raise NotImplementedError("Base class")
 
 
 class TaskFuture(_Future):
     """Implementation of Future for tasks and 'ExecutorFutureHandler'"""
 
-    def __init__(self, task_exec_id: str, executor: ExecutorFutureHandler, **kwargs):
+    def __init__(
+        self, task_exec_id: TaskExecutionID, executor: ExecutorFutureHandler, **kwargs
+    ):
         super().__init__()
+        if not isinstance(executor, ExecutorFutureHandler):
+            raise TypeError
 
         self._executor = weakref.ref(executor)
         self.task_kwargs = {}
--- a/src/ewoksorange/gui/concurrency/_future.py
+++ b/src/ewoksorange/gui/concurrency/_future.py
@@ -1,4 +1,3 @@
-import logging
 import weakref
 from concurrent.futures import Future as _Future
 from typing import TYPE_CHECKING
@@ -6,13 +5,16 @@ from typing import TYPE_CHECKING
 if TYPE_CHECKING:
     from .base import TaskExecutionID
 
-_logger = logging.getLogger(__name__)
-
 
 class ExecutorFutureHandler:
     """Define internal API to cancel a future."""
 
-    def _cancel_future(self, future: TaskFuture) -> None:
+    def _cancel_future(self, future: TaskFuture) -> bool:
+        """Cancel a pending future"""
+        raise NotImplementedError("Base class")
+
+    def _abort_future(self, future: TaskFuture) -> bool:
+        """Abort a running future"""
         raise NotImplementedError("Base class")
 
 
@@ -35,9 +37,16 @@ class TaskFuture(_Future):
         return self._executor()
 
     def cancel(self) -> bool:
+        # The Future 'cancel' API only works if the start hasn't cancel yet...
         if self.done():
             return False
-        if self.executor:
-            self.executor._cancel_future(self)
-
+        if not self.executor:
+            return False
+        self.executor._cancel_future(self)
         return super().cancel()
+
+    def abort(self) -> bool:
+        if not self.executor:
+            return False
+        res = self.executor._abort_future(self)
+        return res
--- a/src/ewoksorange/gui/concurrency/_future.py
+++ b/src/ewoksorange/gui/concurrency/_future.py
@@ -1,39 +1,28 @@
+from typing import TYPE_CHECKING
 import weakref
 from concurrent.futures import Future as _Future
-from typing import TYPE_CHECKING
 
 if TYPE_CHECKING:
     from .base import TaskExecutionID
-
-
-class ExecutorFutureHandler:
-    """Define internal API to cancel a future."""
-
-    def _cancel_future(self, future: TaskFuture) -> bool:
-        """Cancel a pending future"""
-        raise NotImplementedError("Base class")
-
-    def _abort_future(self, future: TaskFuture) -> bool:
-        """Abort a running future"""
-        raise NotImplementedError("Base class")
 
 
 class TaskFuture(_Future):
     """Implementation of Future for tasks and 'ExecutorFutureHandler'"""
 
     def __init__(
-        self, task_exec_id: TaskExecutionID, executor: ExecutorFutureHandler, **kwargs
+        self,
+        task_exec_id: "TaskExecutionID",
+        executor,
+        **kwargs,
     ):
         super().__init__()
-        if not isinstance(executor, ExecutorFutureHandler):
-            raise TypeError
 
         self._executor = weakref.ref(executor)
         self.task_kwargs = {}
         self.task_exec_id = task_exec_id
 
     @property
-    def executor(self) -> ExecutorFutureHandler | None:
+    def executor(self):
         return self._executor()
 
     def cancel(self) -> bool:
--- a/src/ewoksorange/gui/concurrency/base.py
+++ b/src/ewoksorange/gui/concurrency/base.py
@@ -7,6 +7,8 @@ from typing import TypeAlias
 from ewokscore import TaskWithProgress
 from ewokscore.task import Task
 from ewokscore.task import TaskInputError
+from ._future import TaskFuture
+from concurrent.futures import InvalidStateError
 
 _logger = logging.getLogger(__name__)
 
@@ -35,24 +37,30 @@ class TaskExecutor:
             else:
                 _logger.info(f"task initialization failed: {e}")
 
-    def execute_task(self) -> TaskExecutionID:
+    def execute_task(self) -> TaskFuture:
         """
         Execute the task and return a tuple indicating success of the submission and the execution ID.
         """
-        if not self.has_task:
-            return str(
+        future = TaskFuture(
+            task_exec_id=str(
                 uuid.uuid4()
-            )  # Return a dummy TaskExecutionID since we couldn't create the task
+            ),  # Use a dummy TaskExecutionID since we couldn't create the task)
+            executor=self,
+        )
+        if not self.has_task:
+            # if no task defined this mean that initialization has failed.
+            future.set_exception(InvalidStateError("Task not defined."))
+            return future
+
         try:
             self.__task.execute()
         except Exception as e:
             _logger.error(f"task failed: {e}", exc_info=True)
-            return str(
-                uuid.uuid4()
-            )  # Return a dummy TaskExecutionID since we couldn't create the task
-        return str(
-            uuid.uuid4()
-        )  # Return a dummy TaskExecutionID since we couldn't create the task
+            future.set_exception(exception=e)
+        else:
+            # To be discussed. What should be the result ? output_values
+            future.set_result(self.__task.output_values)
+        return future
 
     @property
     def has_task(self) -> bool:
--- a/src/ewoksorange/gui/concurrency/base.py
+++ b/src/ewoksorange/gui/concurrency/base.py
@@ -59,7 +59,9 @@ class TaskExecutor(ExecutorFutureHandler):
             _logger.error(f"task failed: {e}", exc_info=True)
             future.set_exception(exception=e)
         else:
-            future.set_result(self.__task.output_values)
+            if not self.__task.cancelled:
+                # TODO: investigate, this enter in conflict in the case of the "TaskExecutorQueue" setting result is called twice.
+                future.set_result(self.__task.get_output_values())
         return future
 
     @property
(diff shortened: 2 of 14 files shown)


src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -317,7 +317,7 @@ def task_executor_queue(self):
         """Access the underlying TaskExecutorQueue."""
         return self.__task_executor_queue
 
-    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None:
+    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> str:
```
payno: This is one of the debatable part of the PR is that now `_execute_ewoks_task` and `execute_ewoks_task_without_propagation` now return the task id to be able to cancel it downstream.

But this was the simpler to match the current API. An alternative could be to let the user providing this identifier but this make less sense in point of view.

woutdenolf: Can you show how this would be used? If we need it I would do it for all widgets for consistence.

payno:
Related to https://gitlab.esrf.fr/tomotools/tomwer/-/merge_requests/1429

src/ewoksorange/gui/owwidgets/threaded.py
payno: # Use Case
In our tomography workflow, certain tasks require GPU acceleration. To manage GPU concurrency, we use Stacks (a design choice that, I admit, is debatable). With Stacks, multiple volumes can be queued and await processing.
However, volume reconstruction can be time-consuming—anywhere from a few minutes to over an hour. If a user changes their mind or inputs incorrect parameters, they should be able to cancel processing, whether it’s the ongoing task or another queued one.

# How It Works Today
We currently use the "processview" project, which allows registering tasks (referred to as processes in this project). Each process can expose a cancel() API for cancellation. This requires another processview object called a dataset.
In tomwer, datasets are categorized as either "scan/data" or "volume".

# Integration with ewoksorange API
With the ewoksorange API, the client will still need to handle the mapping between the dataset and the ewoksorange task_id. However, I believe this approach is simpler and more powerful, as it allows the client to implement flexible stack management tailored to their needs.

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -606,21 +606,26 @@ def task_output_changed(self) -> None:
         """
         pass
 
-    def execute_ewoks_task(self, log_missing_inputs: bool = True) -> None:
+    def execute_ewoks_task(self, log_missing_inputs: bool = True) -> None | str:
         """
         Execute the Ewoks task and propagate downstream on completion.
 
         :param log_missing_inputs: Whether missing inputs should be logged.
+        :return: Optional task identifier if submitter to a TaskExecutorQueue else None.
```
woutdenolf: I would return the task_id for all widgets and have `cancel_task` and `cancel_all_tasks` as abstract functions to be implemented (when possible) for each base widget.

woutdenolf (changes requested):

src/ewoksorange/gui/owwidgets/base.py
payno: Sounds good to me 👍. I will try to have a look (if not today on Thursday).

payno: @woutdenolf 
I had a quick look for providing an abstract API for it:

The only point to clarify I see is about  ̀OWEwoksWidgetOneThread` / `OWEwoksWidgetNoThread` classes.

* execution fails if a task is already being processed (which make sense).
* In this case the only task we can cancel is the on-going but I from my point of view the API is still valid and makes sense.

So the execution should return a tuple (succeeded, task execution id) to have complete knowledge.
```python
def execute_ewoks_task(self, log_missing_inputs: bool = True) -> (bool, str):
```
And then executions (submission) fails either we return None or an empty string. Depending on the convention we want. Would it be fine with you ?

payno: The second point  I see is that API for `cancel_running_task` doesn't make really sense for the `OWEwoksWidgetOneThreadPerRun` is would be `cancel_running_tasks` but which make less sense for the others...

woutdenolf: Difference between "cancel" and "abort":

- cancel a pending task (does not work when the task is running or has finished): `cancel_task`, `cancel_all_tasks`
- abort a running task (does not work when the task is pending or has finished): `abort_task`, `abort_all_tasks`

Ideally `execute_ewoks_task` would return the task execution id only. If someone would need to know the state, that could be another function.

payno: Fine for returning only the ID.
For the naming convention even if I agree in the case of the stack or the processing on a single thread I guess we want to have a single entry point for cancelling or aborting. Because users will not know if the processing is pending or running.

So either we take one or the other either we found another name...

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -18,28 +30,44 @@ class TaskExecutorQueue(QObject, Queue):
...
+        self._current_task_id: str | None = None
+        self._task_executor: ThreadedTaskExecutor = _ThreadedTaskExecutor(
+            ewokstaskclass=ewokstaskclass
+        )
         self._task_executor.finished.connect(self._process_ended)
-        self._available = True
-        """Simple thread to know if we can do some processing
-        and avoid to mix thinks with QSignals and different threads
-        """
+        self._available: bool = True
```
payno: I would be in favor of removing this variable 'because we should always be able to deduce it from the  `_current_task_id` I think. `self._available == self._current_task_id is None`
And we could create a property
```python
@property
def isAvailable(self):
    return self._current_task_id is None
```
But maybe this is less clear like that ? I guess it depends on people. I can update it if needed.

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -18,28 +30,44 @@ class TaskExecutorQueue(QObject, Queue):
 
     def __init__(self, ewokstaskclass):
         super().__init__()
-        self._task_executor = _ThreadedTaskExecutor(ewokstaskclass=ewokstaskclass)
+        self._task_queue: Deque[str] = deque()
```
payno: We switch from `Queue` to `deque` in order to easily remove any element.

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -54,36 +82,70 @@ def _process_ended_direct(self, task_executor: "_ThreadedTaskExecutor"):
...
         self._task_executor.stop(wait=True)
         self._task_executor = None
 
     @property
     def current_task(self):
         return self._task_executor.current_task
 
+    def __len__(self):
+        return len(self._task_queue) + (1 if self._current_task_id else 0)
+
```
payno: This was added to get closer to the initial `queue` API. The 'issue' is that we are providing (public) API to retrieve those executor (`OWEwoksWidgetWithTaskStack.task_executor_queue` ) and I think this is API users might us. But I am also fine not having it.

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -46,6 +49,9 @@ def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None
                 self.propagate_downstream()
         finally:
             self._output_changed()
+        return str(
+            uuid.uuid4()
+        )  # Return a dummy TaskExecutionID since we're synchronous
```
payno: This is a bit weird to return an identifier that could be anything and has no utility because execution will already be done.
I considered returning an empty string but for other widgets it will be convention when 'submission' failed. So with it we fulfill the API... but maybe there is room for improvement.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -273,7 +306,16 @@ def __remove_task_executor(self, task_executor: ThreadedTaskExecutor):
...
+        return self.__task_executors.get(id(task_executor), (None, False, ""))[1]
+
+    def _get_task_executor(
+        self, task_id: TaskExecutionID
+    ) -> ThreadedTaskExecutor | None:
+        """Return outputs from the last finished task executor."""
+        task_id_to_executor = {
+            task_id: executor for executor, _, task_id in self.__task_executors.values()
+        }
+        return task_id_to_executor.get(task_id, None)
```
payno: Retrieving the executor forces each time to recreate this `task_id_to_executor` which is far from being perfect.
Nevertheless I don't think this will be call so often and worth adding complexity and some refactoring.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -291,6 +333,17 @@ def _get_task_outputs(self) -> dict:
...
+        raise NotImplementedError
+
+    def cancel_task(self, task_id):
+        self._cancel_running_task(task_id)
+
+    def _cancel_running_task(self, task_id):
+        executor = self._get_task_executor(task_id)
+        if executor is not None:
+            executor.cancel_running_task()
+
```
payno: Having this `_cancel_running_task` was to match the discussion you had. But reading it back in this case this doesn't make much sense.

src/ewoksorange/tests/test_OneThreadPerRun.py
```diff
@@ -65,3 +79,37 @@ def test_OWEwoksWidgetOneThreadPerRun(qtapp):
     values = [obj.value for obj in objects]
     expected = [0, 1, 2]
     assert values == expected
+
+
+def test_OWEwoksWidgetOneThreadPerRun_cancellation(qtapp):
```
payno: I was not very inspired. This is only testing the `cancel_task` and looks very close to `test_OWEwoksWidgetOneThreadPerRun`... But not sure we can win much on grouping the two tests (or testing the `cancel_all_tasks`.

src/ewoksorange/gui/concurrency/base.py
```diff
@@ -31,13 +34,18 @@ def create_task(self, log_missing_inputs: bool = False, **kwargs) -> None:
...
+        """
         if not self.has_task:
-            return
+            return ""
         try:
             self.__task.execute()
         except Exception as e:
             _logger.error(f"task failed: {e}", exc_info=True)
+            return ""
+        return ""
```
payno: At the moment I am still hesitant between returning an empty string or a random value which as no much sense...

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -54,36 +82,71 @@ def _process_ended_direct(self, task_executor: "_ThreadedTaskExecutor"):
...
         """
-        if not self.is_available:
-            with block_signals(self._task_executor):
-                self._task_executor.cancel_running_task()
-                # stop and remove the current task from the stack
-                self._task_executor.stop(wait=wait)
-                # signal that processing is done
-                self._process_ended_direct(task_executor=self._task_executor)
+        if self.is_available:
+            return
```
payno, line 124: I inverted the `if not self.is_available:` for readability. A bit borderline I admit.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -111,27 +121,35 @@ def __init__(self, *args, **kwargs):
...
-    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None:
+    def _execute_ewoks_task(
+        self, propagate: bool, log_missing_inputs: bool
+    ) -> TaskExecutionID:
         """
         Prepare and start the background thread if idle.
 
         :param propagate: Whether to propagate outputs after execution.
         :param log_missing_inputs: Whether to log missing input warnings.
+        :return: Tuple indicating if the task was successfully queued and its execution ID.
```
woutdenolf: Not a tuple

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -186,40 +211,49 @@ def __init__(self, *args, **kwargs):
...
-    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None:
+    def _execute_ewoks_task(
+        self, propagate: bool, log_missing_inputs: bool
+    ) -> TaskExecutionID:
         """
         Create a fresh ThreadedTaskExecutor, register it, and start it if it has work.
 
         :param propagate: Whether to propagate outputs after execution.
         :param log_missing_inputs: Whether to log missing input warnings.
+        :return: Tuple indicating if the task was successfully queued and its execution ID.
```
woutdenolf: Not a tuple

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -606,21 +607,26 @@ def task_output_changed(self) -> None:
...
+        return self._execute_ewoks_task(
+            propagate=True, log_missing_inputs=log_missing_inputs
+        )
 
-    def execute_ewoks_task_without_propagation(self) -> None:
+    def execute_ewoks_task_without_propagation(self) -> TaskExecutionID:
         """
         Execute the Ewoks task without propagating outputs downstream.
+
+        :return: Task identifier if execution was queued, None otherwise.
```
woutdenolf: Does not match typing

woutdenolf (changes requested):
An task id which is an empty string means failure? I'm not in favor. As task id is a task id, regardless of the state of the task. Why do we need this?

src/ewoksorange/gui/concurrency/base.py
```diff
@@ -75,3 +90,9 @@ def output_variables(self) -> Optional[dict]:
     @property
     def current_task(self) -> Optional[Task]:
         return self.__task
+
+    def cancel_task(self, task_exec_id: TaskExecutionID) -> None:
```
payno: I was hesitating between on the naming of this parameter between `task_id` and `task_exec_id`. The first one is simpler but I went for the second one because it appears clearer and avoid confusion with `task_identifier` used in the workflow definition.

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -18,28 +30,42 @@ class TaskExecutorQueue(QObject, Queue):
...
+        self._task_queue: Deque[TaskExecutionID] = deque()
+        """Queue storing task IDs in FIFO order"""
+        self._task_exec_ids: Dict[TaskExecutionID, Dict[str, Any]] = {}
+        """Dictionary mapping task IDs to their keyword arguments"""
+        self._current_task_exec_id: TaskExecutionID | None = None
+        self._task_executor: ThreadedTaskExecutor = _ThreadedTaskExecutor(
+            ewokstaskclass=ewokstaskclass
+        )
         self._task_executor.finished.connect(self._process_ended)
-        self._available = True
```
payno, line 23: It has been replaced by `_current_task_exec_id`. 
The processing is available when there is no current task execution.

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -1,15 +1,27 @@
...
 
 from ..qt_utils.signals import block_signals
+from .base import TaskExecutionID
 from .threaded import ThreadedTaskExecutor
 
 
-class TaskExecutorQueue(QObject, Queue):
-    """Processing Queue with a First In, First Out behavior"""
+class TaskExecutorQueue(QObject):
+    """
```
payno, line 24: This is breaking the API. Because we are no more inheriting from `Queue`.

src/ewoksorange/gui/concurrency/base.py
```diff
@@ -75,3 +90,9 @@ def output_variables(self) -> Optional[dict]:
...
+        pass
+
+    def cancel_all_tasks(self) -> None:
+        pass
```
payno: This is a think I don't really like in this PR. As `TaskExecutor` is not a base class we need to implement those two functions and on the case of task cancellation there is no much we can do on this case...
We could imagine alternatives like creating a real base class but nothing really worthit IMO.

src/ewoksorange/gui/concurrency/_future.py
```diff
@@ -1,23 +1,32 @@
...
-from .base import TaskExecutionID
+from concurrent.futures import Future as _Future
+from typing import TYPE_CHECKING
 
-_logger = logging.getLogger(__name__)
+if TYPE_CHECKING:
+    from .base import TaskExecutionID
 
 
 class ExecutorFutureHandler:
```
woutdenolf: Confusing name: is it the _executor_ or the _future_?

woutdenolf: For reference, this is the typical interface of an _executor_ and a _future_

```python
from abc import ABC, abstractmethod
from enum import Enum, auto


class FutureState(Enum):
    PENDING = auto()
    RUNNING = auto()
    FINISHED = auto()
    CANCELLED = auto()


class Future(ABC):
    """Abstract Future API."""

    @abstractmethod
    def cancel(self) -> bool:
        """Attempt to cancel the task."""
        pass

    @abstractmethod
    def cancelled(self) -> bool:
        """Return True if the task was cancelled."""
        pass

    @abstractmethod
    def running(self) -> bool:
        """Return True if the task is currently running."""
        pass

    @abstractmethod
    def done(self) -> bool:
        """Return True if the task has completed."""
        pass

    @abstractmethod
    def result(self, timeout=None):
        """Return the task result or raise its exception."""
        pass

    @abstractmethod
    def exception(self, timeout=None):
        """Return the exception raised by the task."""
        pass

    @abstractmethod
    def add_done_callback(self, fn):
        """Register a callback invoked when the future completes."""
        pass


class Executor(ABC):
    """Abstract Executor API."""

    @abstractmethod
    def submit(self, fn, /, *args, **kwargs) -> Future:
        """Schedule fn(*args, **kwargs) and return a Future."""
        pass

    @abstractmethod
    def shutdown(self, wait=True, *, cancel_futures=False):
        """Clean up executor resources."""
        pass

    def map(self, fn, *iterables, timeout=None, chunksize=1):
        """
        Default implementation similar to concurrent.futures.Executor.map.
        """
        futures = [
            self.submit(fn, *args)
            for args in zip(*iterables)
        ]

        for future in futures:
            yield future.result(timeout=timeout)

    def __enter__(self):
        return self

    def __exit__(self, exc_type, exc_val, exc_tb):
        self.shutdown(wait=True)
```

payno:
close to the benefit of https://github.com/ewoks-kit/ewoksorange/pull/404/changes and https://github.com/ewoks-kit/ewoksorange/pull/403 that still need a bit of work at the moment.

PR #391, opened 2026-06-02
refactor[ParameterForm]: Udpate`addParameter` docstring; create an enum for `select` values

maxenceprog:


src/ewoksorange/gui/widgets/parameter_form.py
```diff
@@ -63,7 +77,7 @@ def addParameter(
         label: Optional[str] = None,
         readonly: Optional[bool] = None,
         enabled: Optional[bool] = None,
-        select: Optional[str] = None,
+        select: Optional[SelectMode] = None,
```
woutdenolf: This would break lots of projects. Perhaps we can do `select: Union[SelectMode,str,None] = None`

```python
try:
    from enum import StrEnum
except ImportError:
    from backports.strenum import StrEnum


class SelectMode(StrEnum):
```

which requires

```python
dependencies = [
    ...,
    "backports.strenum;python_version < '3.11'",
]
```

and in the function

```
if select is not None:
    select = SelectMode(select)
```

So then you have validation with the enum but you  can still provide a string.

What do you think?

maxenceprog: yes that's a good idea ! I'll do that

maxenceprog:
Is it good for you @woutdenolf ?

woutdenolf (approved):
Thanks!

PR #392, opened 2026-06-02
feat[ParameterForm]: #36 Add an auto-completer for file / directory selection

### LINKED ISSUE #36: QParameterForm: add auto completion to input for fie system path
***In GitLab by @payno on Mar 28, 2024, 11:19 GMT+1:***

It would be nice to add autocompletion for file and simplify users life. Here is an example of such usage:
```python

self.completer = qt.QCompleter()
model = qt.QDirModel(self.completer)
if filters is not None:
    model.setFilter(filters)
self.completer.setModel(model)
self.setCompleter(self.completer)

```
See another example in https://gitlab.esrf.fr/tomotools/tomwer/-/blob/main/tomwer/gui/qlefilesystem.py?ref_type=heads

*Migrated from GitLab: https://gitlab.esrf.fr/workflow/ewoks/ewoksorange/-/issues/36*

maxenceprog:
close #36

### DIFF
--- a/src/ewoksorange/gui/widgets/parameter_form.py
+++ b/src/ewoksorange/gui/widgets/parameter_form.py
@@ -18,6 +18,7 @@ from typing import Union
 from AnyQt import QtCore
 from AnyQt import QtWidgets
 from ewokscore import missing_data
+from silx.gui import qt
 from silx.gui.dialog.DataFileDialog import DataFileDialog
 
 from ..qt_utils.signals import block_signals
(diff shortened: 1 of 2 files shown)


src/ewoksorange/gui/widgets/parameter_form.py
```diff
@@ -703,6 +713,16 @@ def _list_value_first(self, name: str) -> Optional[str]:
...
+    def _attach_path_completer(
+        self, widget: QtWidgets.QLineEdit, dir_only: bool = False
+    ) -> None:
+        completer = QtWidgets.QCompleter(widget)
+        model = QtGui.QFileSystemModel(completer)
+        if dir_only:
+            model.setFilter(QtCore.QDir.AllDirs | QtCore.QDir.NoDotAndDotDot)
+        completer.setModel(model)
+        widget.setCompleter(completer)
+
```
payno, line 732: Looks good, nevertheless I recall we use to have some troubles at esrf with PyQt5.

See https://gitlab.esrf.fr/XRD/darfix/-/merge_requests/410/diffs.

This is probably worth checking if this is still the case (and if this is the case integrate the workaround done :( )

maxenceprog: how can i check that ? Just have a try with pyqt5 ?

payno: Try this code on the cluster and you will see ;)

maxenceprog: I guess it was before cluster update to ubuntu 24 ? I did'nt see anything. Just in case i use qt.QDirModel

maxenceprog:
close because i need to merge in main

PR #393, opened 2026-06-02
fix[ParameterForm]: `QDialog.exec_` is not available in Qt6

maxenceprog:
Fix exception in Qt6 : `QDialog.exec_` method does not exist


woutdenolf:
Does `exec()` exist for PyQt5? Many apps are still using that.

payno (approved):
Nice catch thanks 👍

maxenceprog:
> Does `exec()` exist for PyQt5? Many apps are still using that.

Yes this works in qt5 but removed in qt6

PR #395, opened 2026-06-02
My-first-widget tutorial: split materials

payno:
This PR generalize splitting of materials into three files.

* one for the task
* one for the widget
* one for the orange add-on

The idea is to make it clearer to the user that those should be split, simpler to read and for us easier to maintain :crossed_fingers: (like doing diff between each step)

**Note:** : https://github.com/ewoks-kit/ewoksorange/pull/396 to be reviewed first


doc/tutorials/my_first_widget/display_output.rst
```diff
@@ -13,6 +13,10 @@ First let's move the 'MyWidget' to the control area
 From orange main area to control area
 """""""""""""""""""""""""""""""""""""
 
+.. comment::
+
+    for the following code-block this is not the full diff. But only a subset. This is why we are force to have it in "raw".
```
loichuder:
```suggestion
    The following code-block is only a subset of the full diff. This is why it is displayed in raw format.
```

loichuder (approved):
Now that files are split, the duplication is really clear.

Do you think it could be possible to reduce it? This would be a future PR: this PR is fine since the duplication was already there.

payno:
> Now that files are split, the duplication is really clear.
> 
> Do you think it could be possible to reduce it? This would be a future PR: this PR is fine since the duplication was already there.

Good question thanks. Indeed the `clipdata.py` file is the same the entire exercise so we could consider not duplicating it.
But having it once could interrogate. We could also use some symlink to reduce space but the day we want to modify it might also interrogate.
So I went for the simplest. But I don't have strong opinion on it either. If you are in favor of one of the two solutions (or a third one) we could implement it.

PR #403, opened 2026-06-24
Split multithreaded executor

payno:
This PR extract a multiprocessing executor from the `OWEwoksWidgetOneThreadPerRun` class.

The goal is to have a clearer separation between the widget and the executor and ease refactoring proposed in https://github.com/ewoks-kit/ewoksorange/pull/404

Note: This PR was done using Claude.

PR #405, opened 2026-06-29
add `ewoksorange.gui.concurrency.executor` module

payno:
This PR; done with Claude support add a `gui.concurrency.executor` module.

This module include mechanism for execution in the context of ewoks tasks.

The goal is to replace the existing widget executors by this module and add a "demo" for the multi-process. Update of existing will be done in another dedicated PR.

Preparatory PR for https://github.com/ewoks-kit/ewoksorange/pull/406

AI disclosure: Done using Claude


src/ewoksorange/tests/test_executor.py
```diff
@@ -0,0 +1,218 @@
...
+    params=[
+        pytest.param(
+            (ThreadPoolExecutor, 1, SubmitPolicy.DROP_IF_BUSY), id="thread-1-drop"
+        ),
+        pytest.param((ThreadPoolExecutor, 1, SubmitPolicy.ALWAYS), id="thread-1-queue"),
+        pytest.param(
+            (ThreadPoolExecutor, 4, SubmitPolicy.ALWAYS), id="thread-4-parallel"
+        ),
+        pytest.param(
+            (ProcessPoolExecutor, 2, SubmitPolicy.ALWAYS), id="process-2-parallel"
```
payno, line 65: I only added one test with the `ProcessPoolExecutor` as mentioned in the code this is more some code to show the way to go but most likely some time will be needed to check possible bottlenecks.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -0,0 +1,194 @@
...
+            # SyncManager server process provides proxy objects that are
+            # picklable and safe to pass through ProcessPoolExecutor's pickle
+            # serialisation.  A plain multiprocessing.Queue is NOT picklable.
+            self._manager = multiprocessing.Manager()
+        return self._manager
+
+    def _submit_process(self, task_class, task_kwargs) -> TaskFuture:
+        manager = self._get_manager()
+        started_queue = manager.Queue()
+        abort_event = manager.Event()
```
payno, line 120: At the time each time we submit a task we will need the creation of those two objects (`manager.Queue` and `manager.Event`) that from what I have on mind can take a bit of time.
I think in general as written on the doc the `ProcessPool` is more a prototype in order to 'show the way'.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -0,0 +1,194 @@
...
+
+        callable_obj = _ProcessCallable(
+            task_class, task_kwargs, started_queue, abort_event
+        )
+        worker = _EwoksProcessWorker(abort_event)
+        self_ref = weakref.ref(self)
```
payno, line 127: used to emit the requested signal downstream

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -0,0 +1,194 @@
...
+        _holder: list = [None]
```
payno, line 128: small hack to get the task future defined downstream and needed at execution time.

src/ewoksorange/gui/concurrency/executor/_EwoksThreadHandle.py
```diff
@@ -0,0 +1,41 @@
...
+
+
+class EwoksThreadWorker:
+    """Callable that instantiates and executes an ewoks task in the worker thread."""
+
+    def __init__(self, task_class: Type[Task], **task_kwargs):
+        self._task_class = task_class
+        self._task_kwargs = task_kwargs
+        self._task: Optional[Task] = None
+        self._lock = Lock()
```
payno, line 18: It makes sure there is no conflict when `abort` is called (if _task is not created yet in __call__).

src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
```diff
@@ -0,0 +1,61 @@
...
+
+from ewokscore.task import Task
+
+# Picklable substitute for ewokscore Variable — exposes .value so that the
+# same ``{k: v.value for k, v in result.items()}`` pattern works for both
+# thread and process executors.
+_Var = namedtuple("_Var", ["value"])
+
+
+class ProcessCallable:
```
payno, line 7: This part is needed for submitting process only.

src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
```diff
@@ -0,0 +1,61 @@
...
+        task_kwargs: dict,
+        started_queue,
+        abort_event,
+    ):
+        self._task_class = task_class
+        self._task_kwargs = task_kwargs
+        self._started_queue = started_queue
+        self._abort_event = abort_event
+
+    def __call__(self):
```
payno, line 31: This is the implementation proposed by Claude. The combination of thread / process seems a bit heavy.

An alternative could be to say that processes are not cancelled to not use the task.cancel and fill directly the process (using process.kill).

But again I think this submission to process is more a demo at the moment and to be modified by people concerned about this use case as they will probably have specific use cases.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -0,0 +1,194 @@
...
+
+    Warning: the `ProcessPoolExecutor` implementation is a beta-version. Please.
+    """
+
+    submitted = Signal(object)
+    started = Signal(object)
+    succeeded = Signal(object, object)
+    failed = Signal(object, object)
+    ignored = Signal()
+    finished = Signal(object)
```
payno, line 58: An alternative could be to return the finished with a state (succeeded or failed) but this won't make sense to return the results or exception.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -0,0 +1,194 @@
...
+            return
+
+        try:
+            result = raw_future.result()
+        except Exception as exc:
+            self.failed.emit(task_future, exc)
+        else:
+            self.succeeded.emit(task_future, result)
+        finally:
+            self.finished.emit(task_future)
```
payno, line 193: Maybe this part could be simplified by only emitting `finished` and the future. From this future we can retrieve the exception or result.

woutdenolf:
Oef, much more complex than my proposal https://github.com/ewoks-kit/ewoksorange/pull/403#issuecomment-4818740846

In the spirit of "separation of concerns" could be make the executor Ewoks agnostic?

Can you list the pro's of this approach with respect to https://github.com/ewoks-kit/ewoksorange/pull/403#issuecomment-4818740846 ?

If we can't use the existing 

```python
from concurrent.futures import ProcessPoolExecutor
from concurrent.futures import ThreadPoolExecutor
```

can we at least derive executors from

```python
from concurrent.futures import Executor
```

so we don't have to invent things like `TaskFuture`?

Could you also add a standalone example similar to this one in the scripts folder?

https://github.com/ewoks-kit/ewoksorange/pull/403#issuecomment-4818740846

payno:
Well I think the key point is that we want to be able to call the ewoks task `cancel` function in order to be able to abort it (sorry for misleading names).

And this is why we have this `TaskFuture` and the `EwoksExecutor` which is wrapping the a `concurrent.futures.Executor`.

Behind it we use it as:
```python
        EwoksExecutor(
            ThreadPoolExecutor(max_workers=self._MAX_WORKERS),  # or ProcessPoolExecutor
            self._SUBMIT_POLICY,
        )
```
Then the Process iteration also add some complexity because it requires the `ProcessCallable`. Even if on this one (as mentioned in the PR) maybe there is a way to simplify things by calling `process.kill ` directly.

Here is the 'gui' updated with the code you had on your PR.

```python
import logging
import sys
from concurrent.futures import ProcessPoolExecutor
from concurrent.futures import ThreadPoolExecutor

from AnyQt.QtWidgets import QApplication
from AnyQt.QtWidgets import QGridLayout
from AnyQt.QtWidgets import QPushButton
from AnyQt.QtWidgets import QTextEdit
from AnyQt.QtWidgets import QVBoxLayout
from AnyQt.QtWidgets import QWidget
from ewokscore.tests.examples.tasks.sumtask import SumTask
from ewoksorange.gui.concurrency.executor import EwoksExecutor, SubmitPolicy


class Window(QWidget):
    def __init__(self):
        super().__init__()

        self.setWindowTitle("EwoksExecutor Demo")

        self.log = QTextEdit(readOnly=True)

        self.executors = {
            "T DROP": EwoksExecutor(
                ThreadPoolExecutor(max_workers=1),
                SubmitPolicy.DROP_IF_BUSY,
            ),
            "T QUEUE": EwoksExecutor(
                ThreadPoolExecutor(max_workers=1),
                SubmitPolicy.ALWAYS,
            ),
            "T PARALLEL": EwoksExecutor(
                ThreadPoolExecutor(max_workers=4),
                SubmitPolicy.ALWAYS,
            ),
            "P DROP": EwoksExecutor(
                ProcessPoolExecutor(max_workers=1),
                SubmitPolicy.DROP_IF_BUSY,
            ),
            "P QUEUE": EwoksExecutor(
                ProcessPoolExecutor(max_workers=1),
                SubmitPolicy.ALWAYS,
            ),
            "P PARALLEL": EwoksExecutor(
                ProcessPoolExecutor(max_workers=4),
                SubmitPolicy.ALWAYS,
            ),
        }

        for name, executor in self.executors.items():
            executor.submitted.connect(
                lambda future, n=name: self.on_submitted(n, future)
            )
            executor.ignored.connect(lambda n=name: self.log.append(f"[{n}] ⚠ Ignored"))
            executor.succeeded.connect(
                lambda future, result, n=name: self.log.append(
                    f"[{n}] ✅ {id(future)} -> { {k: v.value for k, v in result.items()} }"
                )
            )
            executor.failed.connect(
                lambda future, exc, n=name: self.log.append(
                    f"[{n}] ❌ {id(future)} -> {exc}"
                )
            )

        layout = QVBoxLayout(self)
        grid = QGridLayout()

        buttons = [
            ("Submit x5 (Thread Drop)", "T DROP"),
            ("Submit x5 (Thread Queue)", "T QUEUE"),
            ("Submit x5 (Thread Parallel)", "T PARALLEL"),
            ("Submit x5 (Process Drop)", "P DROP"),
            ("Submit x5 (Process Queue)", "P QUEUE"),
            ("Submit x5 (Process Parallel)", "P PARALLEL"),
        ]

        for i, (text, key) in enumerate(buttons):
            button = QPushButton(text)
            button.clicked.connect(
                lambda checked=False, k=key: self.submit_many(self.executors[k])
            )
            grid.addWidget(button, i // 2, i % 2)

        layout.addLayout(grid)
        layout.addWidget(self.log)

        self.counter = 1

    def on_submitted(self, name, future):
        if future is None:
            self.log.append(f"[{name}] Submitted -> None")
        else:
            self.log.append(f"[{name}] Submitted -> {id(future)}")

    def submit_many(self, executor):
        for _ in range(5):
            executor.submit_task(
                SumTask, inputs={"a": self.counter, "b": self.counter, "delay": 2}
            )
            self.counter += 1

    def closeEvent(self, event):
        for executor in self.executors.values():
            executor.shutdown()
        super().closeEvent(event)


if __name__ == "__main__":
    logging.basicConfig(level=logging.INFO)

    app = QApplication(sys.argv)

    window = Window()
    window.resize(900, 500)
    window.show()

    sys.exit(app.exec())
```

payno:
We can remove the execution on Process. This is what add the most complexity (see commit https://github.com/ewoks-kit/ewoksorange/pull/405/changes/e3555ab27316907eb5f6fd4efda6e6ace8bd978d)
Without it this simplify a lot. But as this was more or less a target to have at one point I wanted to have an idea of how we could do this.

woutdenolf:
Ok my bad, I saw a raw threading. Thread and queue and thought we were re-implementing the executor logic from python.

src/ewoksorange/gui/concurrency/executor/_EwoksThreadWorker.py
```diff
@@ -0,0 +1,50 @@
...
+        with self._lock:
+            self._task = task
+
+        try:
+            task.execute()
+        except Exception as exc:
+            # ewokscore wraps run() exceptions in RuntimeError; re-raise the
+            # original so callers see the same type as the nothread path.
+            original = task.exception
+            raise original if original is not None else exc
```
woutdenolf: Remove try-except. The raises exception is task.exception. We should fix `TaskExecutor` accordingly. I can do that in a follow-up.

src/ewoksorange/gui/concurrency/executor/_EwoksThreadWorker.py
```diff
@@ -0,0 +1,50 @@
...
+        """Call the ewoks task's cancel() to stop a running task. Returns True if the task was reached."""
+        with self._lock:
+            task = self._task
+            if task is not None:
+                task.cancel()
+                return True
+            return False
+
+    @property
+    def has_task(self) -> bool:
```
woutdenolf: Is this used somewhere?

src/ewoksorange/gui/concurrency/executor/_EwoksProcessWorker.py
```diff
@@ -0,0 +1,16 @@
...
+
+    def __init__(self, abort_event: multiprocessing.Event):  # type: ignore
+        self._abort_event = abort_event
+
+    def abort(self) -> bool:
+        self._abort_event.set()
+        return True
+
+    @property
+    def has_task(self) -> bool:
```
woutdenolf: Is this used anywhere?

src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
```diff
@@ -0,0 +1,61 @@
...
+        # abort() arriving right after started cannot race with reset_state().
+        self._started_queue.put("started")
+
+        try:
+            task.execute()
+        finally:
+            done.set()
+
+        # Return a picklable representation: {name: _Var(value)}
+        return {k: _Var(v.value) for k, v in task.output_variables.items()}
```
woutdenolf: This is really bad because it kills ewoks output caching.

Why is `Variable` not picklable?

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
```diff
@@ -0,0 +1,36 @@
+from concurrent.futures import Future
+
+
+class TaskFuture:
+    """Wraps a concurrent.futures.Future with ewoks-specific abort support."""
+
+    def __init__(self, raw_future: Future, worker):
```
woutdenolf: Use abstract worker as type annotation here.

src/ewoksorange/gui/concurrency/executor/_EwoksProcessWorker.py
```diff
@@ -0,0 +1,16 @@
...
+
+class EwoksProcessWorker:
+    """Controls a task running in a subprocess via a multiprocessing.Event."""
+
+    def __init__(self, abort_event: multiprocessing.Event):  # type: ignore
+        self._abort_event = abort_event
+
+    def abort(self) -> bool:
+        self._abort_event.set()
+        return True
```
woutdenolf, line 13: Not necessarily `True`

woutdenolf (changes requested):
Thanks for the hard work. Perhaps best to discuss in person because there are some tricky things in the current implementation.

# Script

Could you make a `scripts` folder and add the example?

# Task execution logic

When executing a task without cancel/abort, this happens:

- submit task to executor
- emit "submitted" Qt signal with future are argument
- emit "started" Qt signal with future are argument
- emit either
  - "failed" Qt signal with future and exception as argument
  - "succeeded" Qt signal with future and `task.output_variables` as arguments
- emit "finished" with future as argument

The execution state is wrapped with a `EwoksWorker`. It is intended to be executed once.

The `TaskFuture` has a reference to the native future and the `EwoksWorker`.

The `EwoksWorker` exposes `abort()` which is exposed in turn by `TaskFuture`.

When calling `TaskFuture.cancel()` the task gets dropped from the native executor but we don't has a Qt signal for that. ~Also the implementation of the submision blocks until the task is actually running. That should be fixed.~

Also `TaskFuture.abort()` returns `True` when the task is reached and aborted. This is not the case for threads.

Also we need to handle the case where the task gets aborted before it even gets scheduled but the native executor.

Is `TaskFuture.abort()` supposed to be a blocking call? If not how can it return `True` or `False`, unless there is an "_aborted" state?

# Harmonization

Both `EwoksExecutor._submit_thread` and `EwoksExecutor._submit_process` achieve
the same thing but it is very hard to know from reading the code.

- `EwoksThreadWorker` -> `ProcessCallable` + `EwoksProcessWorker`
- `run` -> `relay_started` + `ProcessCallable`

The logic is valid ~(except that it blocks until the task is running which needs fixing)~ but needs refactoring for readability and harmony.

That sounds like an ideal AI job.

If I would do it manually I would start by

- making an abstract `EwoksWorker`
- rename "worker" to something that reflects what it actually does (this will illustrate what needs to be refactored)
- use that abstract type as a type annotation in `TaskFuture.__init__`
- then the refactoring

woutdenolf (approved):
I made a PR to this PR to add a thread cleanup fix script and an example: https://github.com/ewoks-kit/ewoksorange/pull/411.

I'm still wondering whether the split of `_EwoksThreadHandle` into `_EwoksProcessHandle` and `_ProcessCallable` cannot be done in a cleaner way. Perhaps split `_EwoksThreadHandle` in the same way for consistency reasons. But this is internal API and can be changed at any time. Functionally everything is sound now. LGTM after https://github.com/ewoks-kit/ewoksorange/pull/411.

Thanks for the huge effort!

PR #406, opened 2026-06-30
Makes ewoksorange.gui.owwidgets.thread returning `TaskFuture`

payno:
This PR benefits from the `ewoksorange.gui.concurrency.executor` module.

https://github.com/ewoks-kit/ewoksorange/pull/405 to be merged first.

* widgets from `ewoksorange.gui.concurrency.threaded` now use the executors from `ewoksorange.gui.concurrency.executor`
* `execute_ewoks_task` API now return a  ̀TaskFuture ̀ that we can cancel or abort.

Note to reviewers: as the `ewoksorange.gui.owwidgets.threaded` module contains a lot of modification this is probably simpler to look at the existing code today than on the diff.

Note on AI: spend some time to force Claude do what I wanted :innocent: .


src/ewoksorange/gui/concurrency/executor/_EwoksCompletedWorker.py
```diff
@@ -0,0 +1,9 @@
+class CompletedWorker:
```
payno: This worker has been added for the specific use case of the `OWEwoksWidgetNoThread` and in order to propose an homogeneous API (`_execute_ewoks_task` return a `TaskFuture` whenever possible)

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -39,14 +46,25 @@ def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None
         try:
             self.__task_executor.execute_task()
         except Exception as e:
+            exception = e
             _logger.error(f"task failed: {e}", exc_info=True)
+        else:
+            exception = None
+
```
payno: The exception is needed to fill the '_CompletedWorker' downstream.

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -39,14 +46,25 @@ def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None
...
                 self.propagate_downstream()
         finally:
             self._output_changed()
 
+        raw_future = _ConcurrentFuture()
+        if exception is not None:
+            raw_future.set_exception(exception)
+        else:
+            raw_future.set_result(self.__task_executor.output_variables)
+        return TaskFuture(raw_future, _CompletedWorker())
```
payno: Build a dummy TaskFuture to make sure we have a coherent API.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,282 +68,144 @@ def __onProgressChanged(self, progress: int):
         self.progressBarSet(float(progress))
 
 
-class OWEwoksWidgetOneThread(_OWEwoksThreadedBaseWidget, **ow_build_opts):
-    """
-    Single persistent background thread for task execution.
+class _OWEwoksExecutorWidget(_OWEwoksThreadedBaseWidget, **ow_build_opts):
```
payno, line 45: One of the main advantage of using the executor into a dedicated module is that we can have a widget that can be generalized and from which only the "executor" is changed.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,282 +68,144 @@ def __onProgressChanged(self, progress: int):
...
+        self.__executor.succeeded.connect(self.__on_succeeded)
+        self.__executor.failed.connect(self.__on_failed)
+        self.__propagate_by_future: Dict[TaskFuture, bool] = {}
+
+        # Note: all the following variable to be removed since executor refactoring
+        self.__current_task_future: Optional[TaskFuture] = None
+        self.__last_output_variables: dict = {}
+        self.__last_task_succeeded: Optional[bool] = None
+        self.__last_task_done: Optional[bool] = None
+        self.__last_task_exception: Optional[Exception] = None
```
payno: With the new design I think at one point we could get rid of those. It would simplify things and is more aligned with "expectation".

I started to reduce usage but this is not that straightforward so I just added a warning for future.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,282 +68,144 @@ def __onProgressChanged(self, progress: int):
...
+        )
         return self.__last_task_exception
 
     def _get_task_outputs(self) -> dict:
-        """Return the last finished task's outputs."""
+        warnings.warn(
+            f"'cancel_running_task' is deprecated since 6.0 Please use directly the future provided during task submission.",
+            DeprecationWarning,
+            stacklevel=2,
+        )
```
payno: Those are using internally one of the following variables that we would like to remove in the future:
 * `__current_task_future`
 * `__last_output_variables`
 * `__last_task_succeeded`
 * `__last_task_done`
 * `__last_task_exception`

The best would be to able to deprecate them all and use the  ̀TaskFuture ̀ instead but this is not that direct so I only reduce usage of it when this was direct.
To be considered in the future.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,282 +68,144 @@ def __onProgressChanged(self, progress: int):
...
-    """
-
-    def __init__(self, *args, **kwargs):
-        """
-        Initialize the FIFO TaskExecutorQueue.
-        """
-        super().__init__(*args, **kwargs)
-        self.__task_executor_queue = TaskExecutorQueue(
-            ewokstaskclass=self.ewokstaskclass
+    def cancel_running_task(self) -> None:
```
payno, line 141: We should now use future. Internally we are only using it for test that will be removed once `TaskExecutorQueue` class is removed.

src/ewoksorange/tests/test_task_executor.py
payno, line 1: replaced by `test_executor` (from previous PR)

payno, line 1: Removed as all the classes it was testing are now deprecated. It also avoid having some modifications of the .github/workflows/test.yml to ignore deprecation warning. A bit violent but way simpler.

src/ewoksorange/tests/test_nothread.py
payno, line 1: Added test for `OWEwoksWidgetNoThread` that were not existing. Simple test to make sure the new API using future is working for this widget.

src/ewoksorange/tests/test_OneThreadPerRun.py
```diff
@@ -52,16 +54,29 @@ def test_OWEwoksWidgetOneThreadPerRun(qtapp):
...
+        if test_case == "cancel_futures":
+            widget.set_dynamic_input("sleep_duration", 0.5)
 
         # Start calculation
-        widget.handleNewSignals()
+        futures.append(widget.execute_ewoks_task())
 
+    if test_case == "cancel_futures":
+        # wait for the future to be started. Should be done through a QtEvent
+        sleep(0.2)
```
payno: I was considering using an Event but this is probably overkilled.

pyproject.toml
```diff
@@ -4,7 +4,7 @@ build-backend = "setuptools.build_meta"
 
 [project]
 name = "ewoksorange"
-version = "5.0.1"
+version = "6.0.0"
```
payno: I tried to minimize API break but this is huge modification. So for sure it will break existing behavior under the hood. More honest to consider this will part of a major release.

src/ewoksorange/gui/concurrency/executor/_EwoksCompletedWorker.py
```diff
@@ -0,0 +1,12 @@
+from ._EwoksWorkerBase import EwoksWorkerBase
+
+
+class CompletedWorker(EwoksWorkerBase):
```
payno: Not a big fan of the name. Better ideas welcomed.

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
```diff
@@ -32,7 +32,8 @@ def done(self) -> bool:
         return self._future.done()
 
     def result(self, timeout=None):
-        return self._future.result(timeout=timeout)
+        raw = self._future.result(timeout=timeout)
+        return {k: v.value for k, v in raw.items()}
```
payno: Seems more user friendly to have the dict directly.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
```diff
@@ -153,16 +153,15 @@ def _finalize_submission(
...
-        # calls the callback synchronously in this thread.
-        self.submitted.emit(task_future)
         raw_future.add_done_callback(
             lambda f: self._done_callback(self_ref, f, task_future)
         )
+        self.submitted.emit(task_future)
+        # Register the callback and emit submitted BEFORE _ready.set() so the
+        # task cannot finish before both are in place.
+        if ready is not None:
+            ready.set()
```
payno, line 160: This was triggering some troubles with widget test. But indeed better wait as much as possible before declaring it 'ready'

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -285,93 +121,52 @@ def task_done(self) -> Optional[bool]:
 
     @property
     def task_exception(self) -> Optional[Exception]:
-        return self.__last_task_exception
+        exc = self.__last_task_exception
+        # task.execute() wraps run() exceptions as RuntimeError(...) from original;
+        # follow __cause__ to surface the exception the task actually raised.
+        return exc.__cause__ or exc if exc is not None else None
```
payno, line 132: This one is tricky. Going for checking the __cause__ was the simplest but not very friendly I have to admit. No better solution found at the moment.

src/ewoksorange/gui/concurrency/executor/_EwoksCompletedHandle.py
```diff
@@ -0,0 +1,12 @@
+from ._EwoksTaskHandle import EwoksTaskHandle as _EwoksTaskHandle
+
+
+class EwoksCompletedHandled(_EwoksTaskHandle):
```
loichuder:
```suggestion
class EwoksCompletedHandle(_EwoksTaskHandle):
```

src/ewoksorange/gui/concurrency/executor/_EwoksProcessHandle.py
```diff
@@ -1,10 +1,12 @@
+import multiprocessing
+
 from ._EwoksTaskHandle import EwoksTaskHandle
 
 
 class EwoksProcessHandle(EwoksTaskHandle):
     """Controls a task running in a subprocess via multiprocessing.Event objects."""
 
-    def __init__(self, abort_event, aborted_event):
+    def __init__(self, abort_event: multiprocessing.Event, aborted_event: multiprocessing.Event):  # type: ignore
```
loichuder: Not sure `# type: ignore` is needed. We don't really enforce static typing?

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
loichuder: It is guaranteed that `raw` is a dict?

CHANGELOG.md
loichuder, line 1: `Deprecated` section should be added for the deprecated classes

src/ewoksorange/tests/test_OneThreadPerRun.py
loichuder: Then no reason to keep it in the comment. See suggestion above.

src/ewoksorange/tests/test_OneThreadPerRun.py
```diff
@@ -52,16 +54,29 @@ def test_OWEwoksWidgetOneThreadPerRun(qtapp):
...
         widget.set_dynamic_input("my_object", obj)
+        if test_case == "cancel_futures":
+            widget.set_dynamic_input("sleep_duration", 0.5)
 
         # Start calculation
-        widget.handleNewSignals()
+        futures.append(widget.execute_ewoks_task())
 
+    if test_case == "cancel_futures":
+        # wait for the future to be started. Should be done through a QtEvent
```
loichuder:
```suggestion
        # Wait for the future to start
```

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
woutdenolf: Are these `Variable` objects? If so then this is not the place to dereference them.

src/ewoksorange/tests/test_OneThreadPerRun.py
```diff
@@ -38,7 +27,20 @@ class OWEwoksWidgetOneThreadPerRun(
...
+        (
+            "standard_execution",
+            [0, 1, 2],
+        ),
+        (
+            "cancel_futures",
+            ["cancelled", "cancelled", "cancelled"],
+        ),
+    ],
+)
```
loichuder, line 42: The parallelization is too complex in my opinion making the test difficult to read.

I would suggest to make two tests with code duplication: one for standard execution, the other for cancelation.

Then we can see if some code can be put in common.

src/ewoksorange/gui/concurrency/threaded.py
```diff
@@ -8,6 +9,14 @@
 class ThreadedTaskExecutor(QThread, TaskExecutor):
     """Create and execute an Ewoks task in a dedicated thread."""
```
woutdenolf, line 10:
```suggestion
    """Create and execute an Ewoks task in a dedicated thread.

    .. warning:: deprecated since version 6.0
    """
```

woutdenolf: In analogy with the other one.

src/ewoksorange/tests/test_nothread.py
```diff
@@ -0,0 +1,48 @@
...
+    **ow_build_opts,
+    ewokstaskclass=DummyTask,
+):
+    name = "test_OW"
+
+
+def test_OWEwoksWidgetNoThread(qtapp):
+    """
+    Test processing of two tasks.
+    The first task will be executed, prevent any other execution of code. Then the second will aldo be executed.
```
loichuder:
```suggestion
    The first task will be executed, blocking any other execution of code. Then the second will also be executed.
```

src/ewoksorange/tests/test_nothread.py
```diff
@@ -0,0 +1,48 @@
...
+    _OWEwoksWidgetNoThread,
+    **ow_build_opts,
+    ewokstaskclass=DummyTask,
+):
+    name = "test_OW"
+
+
+def test_OWEwoksWidgetNoThread(qtapp):
+    """
+    Test processing of two tasks.
```
loichuder:
```suggestion
    Test sequential processing of two tasks.
```

src/ewoksorange/gui/owwidgets/threaded.py
loichuder: Should we not surface the full exception stack and let the consumer decide if they want to retrieve the cause?

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -285,93 +121,52 @@ def task_done(self) -> Optional[bool]:
...
-        """
-        Initialize the FIFO TaskExecutorQueue.
-        """
-        super().__init__(*args, **kwargs)
-        self.__task_executor_queue = TaskExecutorQueue(
-            ewokstaskclass=self.ewokstaskclass
+    def cancel_running_task(self) -> None:
+        """Abort the currently running task."""
+        warnings.warn(
+            "'cancel_running_task' is deprecated since 6.0 Please use directly the future provided during task submission.",
```
loichuder:
```suggestion
            "'cancel_running_task' is deprecated since 6.0. Please cancel the task by calling the  `cancel` method of the future provided during task submission.",
```

loichuder (commented):
Again, I focused on the low level stuff. This is not my area of expertise so I can't judge the big picture.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
woutdenolf: We need to ensure the Qt signal order (if not `ignored`):

1. `submitted`
2. `started`
3. `succeeded/failed/aborted`
4. `finished`

Before we had

1. `ready.set`: allow emitting `started` after this
2. `submitted.emit`: emit `submitted`
3. `raw_future.add_done_callback`: emit `succeeded/failed/aborted` when task finished

After this change

1. `raw_future.add_done_callback`: emit `succeeded/failed/aborted` when task finished
2. `submitted.emit`: emit `submitted`
3. `ready.set`: allow emitting `started` after this

I would think this is the proper order

1. `submitted.emit`: emit `submitted`
2. `ready.set`: allow emitting `started` after this
3. `raw_future.add_done_callback`: emit `succeeded/failed/aborted` when task is finished

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,184 +42,74 @@ def __onProgressChanged(self, progress: int):
...
 
-class OWEwoksWidgetOneThread(_OWEwoksThreadedBaseWidget, **ow_build_opts):
-    """
-    Single persistent background thread for task execution.
+class _OWEwoksExecutorWidget(_OWEwoksThreadedBaseWidget, **ow_build_opts):
+    """Base for all EwoksExecutor-backed widgets."""
 
-    A second execution request while the thread is running is refused.
-    """
+    _MAX_WORKERS: Optional[int] = 1
```
woutdenolf: Why is the default 1?

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -26,27 +28,38 @@ def __init__(self, *args, **kwargs):
...
+        self, propagate: bool, log_missing_inputs: bool
+    ) -> Optional[TaskFuture]:
         """
         Create and execute the Task synchronously.
 
         :param propagate: Whether to propagate outputs after execution.
         :param log_missing_inputs: Whether to log missing input warnings.
         """
+        # Both methods handle exceptions internally (ewokscore >= 4.0.1):
+        # create_task() stores TaskInputError silently; execute_task() never raises.
```
woutdenolf: What changed since ewokscore 4.0.1? Raising on execution has always been opt-out afaik

https://github.com/ewoks-kit/ewokscore/blob/1fe98c066398b0a332f4d43ef43f055f125ab504/src/ewokscore/task.py#L739

woutdenolf (changes requested):
I like the direction! I did not fully review `owwidgets` yet.

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
payno: yes. `self._future.result` returns a `VariableContainer`
so `{k: v.value for k, v in raw.items()}` iterates on  ̀Variable` and we dereference them. But from a user point of view when I retrieve the TaskFuture result this is what I expect.

@woutdenolf where would be the place for de-referencing ? You would return the `VariableContainer` directly ? This would force user to know the internal structure to access what they are looking for.

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
payno: At the moment emit of `succeeded/failed/aborted` can never happen before ready is set. as `started`. Because it is waiting over `ready`.

But for me we are 'force' to set 'ready' at the end (after registering the done callback) else if we consider the case of an empty task and concurrency the task could finish (ready released) before the done callback are registered.
And then some signals might not be emit (at the least)

src/ewoksorange/gui/concurrency/executor/_EwoksCompletedHandle.py
payno: :+1: easier to rename with the IDE. d7ec2e723700da080dc6085603735b5eaedc4e9d will fix it.

src/ewoksorange/gui/concurrency/executor/_EwoksProcessHandle.py
payno: Yes, I get a `Variable not allowed in type expression` error if I don't set it, because Event is defined as a function. Since I always import it from multiprocessing directly, I hadn't realized there was a separate class.

For reference: from multiprocessing import Event gives you a function that creates an instance of the class, while the actual class lives in from multiprocessing.synchronize import Event.

Anyway, I've switched to using the class directly (f863c756e52a88c52b47232f604ab129c0838440).

CHANGELOG.md
payno: :+1: 8dac0559a2d7a7e1f4d9c4534c460c7d13fbac45 will fix it.

src/ewoksorange/tests/test_OneThreadPerRun.py
payno: *parallelization -> parameterization I guess.
Not sure I agree to be honest I think we will need a third opinion there :innocent:

src/ewoksorange/tests/test_nothread.py
payno: 299bd8146a60a89c87023307b3f5679d714f84e7

src/ewoksorange/gui/owwidgets/threaded.py
payno: f9bc006a80561f0137aa7b2047966c8e870079d1

src/ewoksorange/gui/owwidgets/threaded.py
payno: To be honest, I asked myself the same question. I don't have a strong opinion on it, but here are the arguments:

* This is the simplest (safest) use case (no concurrency), so it might be easier to redefine another value later.
* There's no reason to add complexity by defining an invalid value and force users to define it. `None` can have side effects for users who aren't aware of what they're doing, and a "random" value (like 4) doesn't make more sense IMO.

If you have other arguments or want to go with a different value, feel free to share your reasoning.

I also asked myself to 'overwrite' for safety in the appropriate widgets and make sure no one will silently modify it and break children classes. But I went for reducing the code. But I can ear a different opinion.

src/ewoksorange/gui/owwidgets/nothread.py
payno: ah indeed...

src/ewoksorange/gui/owwidgets/threaded.py
woutdenolf: The typical default in the number of CPUs no?

src/ewoksorange/gui/concurrency/executor/EwoksExecutor.py
woutdenolf: > could finish (ready released) before the done callback are registered.

`add_done_callback` already has builtin support for that: when the task is finished before adding the callback, calling `add_done_callback` is the same as calling the callback directly. The original comment says so.

src/ewoksorange/gui/concurrency/executor/TaskFuture.py
woutdenolf: Yes we never dereference. Ewokscore does that when needed. See `OWEwoksWidgetNoThread._get_task_outputs`. The type annotation is wrong.

src/ewoksorange/gui/owwidgets/threaded.py
payno: In the past I've also seen just the number of CPUs, or the number of CPUs divided by 2, or number of CPUs plus some fixed value...

I don't have a strong opinion here. We could just define a default value and agree on it, and use `-1` for `OWEwoksWidgetOneThreadPerRun`.
That said, I think this will eventually depend on the use case (e.g. the resource footprint of the task).

woutdenolf: I would say use `None` and leave it up to python's concurrent futures library

https://github.com/python/cpython/blob/aad438b0b61206ff67f32aa1ca24792f26f6fff9/Lib/concurrent/futures/thread.py#L181

https://github.com/python/cpython/blob/22c797005d1eaa2b03bcd9d858c2525a1d1560c4/Lib/concurrent/futures/process.py#L772

That is the natural way.

Then if someone needs limitations (like the queue and the drop), `_MAX_WORKERS` and `_SUBMIT_POLICY`can be modified.

pyproject.toml
woutdenolf: Ok with bumping the major. Typically we do this is a separate PR which then gets released after merging.

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -26,7 +28,9 @@ def __init__(self, *args, **kwargs):
         super().__init__(*args, **kwargs)
         self.__task_executor = TaskExecutor(self.ewokstaskclass)
 
-    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None:
+    def _execute_ewoks_task(
+        self, propagate: bool, log_missing_inputs: bool
+    ) -> Optional[TaskFuture]:
```
woutdenolf: Always returns `TaskFuture`.

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -723,11 +735,14 @@ def __post_task_execute(self, callbacks: List[Callable[[], None]]) -> None:
             if ncallbacks > 1:
                 self.__post_task_execute(callbacks[1:])
 
-    def _execute_ewoks_task(self, propagate: bool, log_missing_inputs: bool) -> None:
+    def _execute_ewoks_task(
+        self, propagate: bool, log_missing_inputs: bool
+    ) -> Optional[TaskFuture]:
```
woutdenolf, line 740: Always returns `TaskFuture`.

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -723,11 +735,14 @@ def __post_task_execute(self, callbacks: List[Callable[[], None]]) -> None:
...
         """
         Subclasses must implement how the task is created and executed.
 
         :param propagate: Whether to propagate outputs downstream after execution.
         :param log_missing_inputs: Whether to log missing input warnings.
+        :return: TaskFuture for async widgets, None for synchronous widgets.
```
woutdenolf: Always returns `TaskFuture`.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,184 +42,74 @@ def __onProgressChanged(self, progress: int):
...
+
+        # Note: all the following variables might be removed in the future. Please avoid using them.
+        self.__current_task_future: Optional[TaskFuture] = None
+        self.__last_output_variables: dict = {}
+        self.__last_task_succeeded: Optional[bool] = None
+        self.__last_task_done: Optional[bool] = None
+        self.__last_task_exception: Optional[Exception] = None
+
+    def _execute_ewoks_task(
+        self, propagate: bool, log_missing_inputs: bool
```
woutdenolf, line 70: `log_missing_inputs` is not passed.

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -96,184 +42,74 @@ def __onProgressChanged(self, progress: int):
...
+    ) -> Optional[TaskFuture]:
```
woutdenolf, line 71: Always returns TaskFuture.

woutdenolf (approved):
Some small fixes but LGTM.

src/ewoksorange/tests/test_OneThreadPerRun.py
woutdenolf: It seems like all tests use this `MyObject` with event pattern. They should be refactored but not in this PR. It has taken way too long already. Lets merge so I can start contributing in cleaning up.

woutdenolf:
Please squash when merging.

src/ewoksorange/gui/owwidgets/threaded.py
woutdenolf: Done

src/ewoksorange/gui/owwidgets/threaded.py
woutdenolf: We can fix in another PR.

src/ewoksorange/gui/owwidgets/base.py
woutdenolf: Ok can be None when dropped. I took the liberty to fix the docstring.

src/ewoksorange/gui/owwidgets/nothread.py
woutdenolf: I took the liberty to fix it.

src/ewoksorange/gui/owwidgets/base.py
woutdenolf: Ok can be None when dropped. I took the liberty to fix the docstring.

src/ewoksorange/gui/owwidgets/threaded.py
woutdenolf: Ok can be None when dropped. I took the liberty to fix the docstring.

pyproject.toml
woutdenolf: I took the liberty to fix it.

payno:
Thanks for the reviews and commits :+1:

src/ewoksorange/gui/concurrency/queued.py
```diff
@@ -18,6 +22,12 @@ class TaskExecutorQueue(QObject, Queue):
 
     def __init__(self, ewokstaskclass):
         super().__init__()
+
+        warnings.warn(
+            "TaskExecutorQueue is deprecated since version 6.0."
+            "Please use 'ewoksorange.gui.concurrency.executor.EwoksExecutor' instead.",
+            DeprecationWarning,
+        )
```
payno: 👍

src/ewoksorange/gui/owwidgets/nothread.py
```diff
@@ -26,27 +28,37 @@ def __init__(self, *args, **kwargs):
...
         try:
-            self.__task_executor.execute_task()
-        except Exception as e:
-            _logger.error(f"task failed: {e}", exc_info=True)
-        try:
-            self.__post_task_exception = None
             if propagate:
-                self.propagate_downstream()
+                # Always pass an explicit bool — passing None triggers a DeprecationWarning
+                # (base.py:533) which becomes an error under pytest -W error.
```
payno: 👍

PR #411, opened 2026-07-16
387 owewokswidgetwithtaskstack provide api to cancel a task 3 add example

woutdenolf:
Add fix and example script

### DIFF
--- a/src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
+++ b/src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
@@ -40,11 +40,17 @@ class ProcessCallable:
         done = threading.Event()
 
         def _watch_abort():
-            self._abort_event.wait()
-            if not done.is_set():
+            try:
+                self._abort_event.wait()
+                if done.is_set():
+                    # abort_event was only set to release this thread.
+                    return
                 task.cancel()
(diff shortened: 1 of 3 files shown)


src/ewoksorange/gui/concurrency/executor/__init__.py
```diff
@@ -1,2 +1,3 @@
 from .EwoksExecutor import EwoksExecutor  # noqa F401
+from .EwoksExecutor import SubmitPolicy  # noqa F401
```
payno, line 2: 👍

src/ewoksorange/gui/concurrency/executor/_ProcessCallable.py
```diff
@@ -57,5 +63,7 @@ def _watch_abort():
             task.execute()
         finally:
             done.set()
+            self._abort_event.set()
```
payno, line 68: I don't see why the abort should always be set. Should it be more upper (line 54 and replacing line 50?)

woutdenolf: this is the stop the `_watch_abort` thread which is waiting on `self._abort_event.wait()`. See above, it has a comment.

payno (approved):

PR #415, opened 2026-07-23
qt_message_handler: fix access to 'QtMsgType' Enum value.

payno:
port https://github.com/ewoks-kit/ewoksorange/pull/414 to 4.1.0 for bug fix release


woutdenolf:
Why is this backport needed?

woutdenolf:
I doubt 5.0.0 would cause problems for you: https://github.com/ewoks-kit/ewoksorange/blob/main/CHANGELOG.md#500---2026-04-13

payno:
Originally because in production. And too late to properly test the 5.0.
But indeed this is a corner case that a but fix release might be over-killed.

PR #419, opened 2026-07-28
Store all graph attributes in OWS format

woutdenolf:
Example:

```bash
ewoks convert demo demo.ows --test  # add more graph attributes
ewoks execute demo.ows --engine orange
```

The workflow description is the place we abuse to store Ewoks graph attributes

image

PR #420, opened 2026-07-28
refactor EwoksExecutor

### LINKED ISSUE #423: Flaky abort test
```
___________________ test_abort_running_task[thread-1-queue] ___________________

executor = <ewoksorange.gui.concurrency.executor.EwoksExecutor.EwoksExecutor object at 0x000001EAD96977F0>

    def test_abort_running_task(executor):
        """abort() exits the task early; succeeded and aborted both fire."""
        started = QtEvent()
        done = QtEvent()
        result_holder = {}
        aborted_holder = {}
    
        executor.started.connect(lambda task_future: started.set())
        executor.succeeded.connect(
            lambda task_future, result: (
                result_holder.update({"result": result}),
                done.set(),
            )
        )
        executor.failed.connect(
            lambda task_future, exception: (
                result_holder.update({"exception": exception}),
                done.set(),
            )
        )
        executor.aborted.connect(
            lambda task_future: aborted_holder.update({"task_future": task_future})
        )
    
        task_future = executor.submit_task(SleepTask, inputs={"duration": 60.0})
        assert started.wait(timeout=5), "task never started"
    
        task_future.abort()
    
>       assert done.wait(timeout=5)
E       assert False
E        +  where False = wait(timeout=5)
E        +    where wait = <ewoksorange.gui.qt_utils.app.QtEvent object at 0x000001EAD960A740>.wait

src\ewoksorange\tests\test_executor.py:140: AssertionError
```

woutdenolf:
Closes #423

This PR:

1. Refactor `EwoksExecutor`: make thread and process implementation equivalent.
    * `TaskRunner`: execute Ewoks task (possibly in sub-process)
    * `TaskController`: currently only `abort`
3. Add synchronous execution support to `EwoksExecutor`: equivalent implementation to thread and process.
    * Will be used in a follow-up PR for the nothread widget, replacing `ewoksorange.concurrency.base.TaskExecutor`
5. Add synchronous execution to `scripts/ewoks_executor_demo.py`
6. Add unit tests for `EwoksExecutor`

The old executor implementation and associated tests are kept in this PR. 

```bash
python scripts/ewoks_executor_demo.py 
[   5.829s] [T QUEUE] Submitted -> 139908628118368
[   5.830s] [T QUEUE] Submitted -> 139908628119088
[   5.831s] [T QUEUE] Submitted -> 139908628251568
[   5.831s] [T QUEUE] Submitted -> 139908628252240
[   5.832s] [T QUEUE] Submitted -> 139908628252912
[   5.832s] [T QUEUE] ▶ Started -> 139908628118368
[   7.832s] [T QUEUE] ✅ 139908628118368 -> {'result': 2}
[   7.881s] [T QUEUE] ■ Finished -> 139908628118368
[   7.881s] [T QUEUE] ▶ Started -> 139908628119088
[   9.833s] [T QUEUE] ✅ 139908628119088 -> {'result': 4}
[   9.834s] [T QUEUE] ■ Finished -> 139908628119088
[   9.834s] [T QUEUE] ▶ Started -> 139908628251568
[  11.834s] [T QUEUE] ✅ 139908628251568 -> {'result': 6}
[  11.835s] [T QUEUE] ■ Finished -> 139908628251568
[  11.835s] [T QUEUE] ▶ Started -> 139908628252240
[  13.835s] [T QUEUE] ✅ 139908628252240 -> {'result': 8}
[  13.836s] [T QUEUE] ■ Finished -> 139908628252240
[  13.836s] [T QUEUE] ▶ Started -> 139908628252912
[  15.836s] [T QUEUE] ✅ 139908628252912 -> {'result': 10}
[  15.836s] [T QUEUE] ■ Finished -> 139908628252912
```

image


woutdenolf:
CI still broken due to `ewokscore` pre-release. PR that fixes it: https://github.com/ewoks-kit/ewoksorange/pull/419.

src/ewoksorange/gui/concurrency/_controllers/abstract.py
```diff
@@ -0,0 +1,19 @@
+from abc import ABC
+from abc import abstractmethod
+
+
+class TaskController(ABC):
+    """Interface for controlling a running ewoks task."""
```
payno: I don't know if this is obvious for everyone. But I might add that this task is expected to be executed one at most.


```suggestion
    """Interface for controlling a running ewoks task. The task (instance) is expected to be executed at most once."""
```

src/ewoksorange/gui/concurrency/_executor/__init__.py
payno, line 1: 👍

src/ewoksorange/gui/concurrency/_runners/process.py
```diff
@@ -0,0 +1,75 @@
...
+from typing import Dict
+from typing import Type
+
+from ewokscore.task import Task
+from ewokscore.variable import VariableContainer
+
+from .abstract import TaskRunner
+
+
+class ProcessTaskRunner(TaskRunner):
```
payno, line 13: Fine with it but I would expect to see the `ProcessCallable` removed. Same purpose no ?

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -0,0 +1,251 @@
...
+from ewokscore.task import Task
+
+from . import _controllers
+from . import _runners
+from .future import TaskFuture
+
+_logger = logging.getLogger(__name__)
+
+
+class SubmitPolicy(Enum):
```
payno, line 27: This is also defined in "ewoksorange.gui.concurrency._executor" if I am correct. Looks like some leftovers are still in the code.

Image

src/ewoksorange/gui/concurrency/future.py
```diff
@@ -0,0 +1,54 @@
+from concurrent.futures import Future
+
+from . import _controllers
+
+
+class TaskFuture:
```
payno, line 6: Duplicated in `concurerncy._executor.TaskFuture` and `concurrency.Future`. I let you check how to solve it. `git mv` issue ?

Image

woutdenolf:
@payno Sorry I was not very clear. The code in `concurrency._executor` will be removed but not in this PR because the nothread widget is till using some of it. In a follow up it will use EwoksExecutor with execute=None.

So yes, there is a decent amount of overlapping code.

src/ewoksorange/gui/concurrency/_runners/process.py
woutdenolf: `ewoksorange.gui.concurrency._executor` is left untouched and will be completely remove when the nothread widget uses EwoksExecutor with execute=None.

src/ewoksorange/gui/concurrency/executor.py
woutdenolf: `ewoksorange.gui.concurrency._executor` is left untouched and will be completely remove when the nothread widget uses EwoksExecutor with execute=None.

src/ewoksorange/gui/concurrency/future.py
woutdenolf: `ewoksorange.gui.concurrency._executor` is left untouched and will be completely remove when the nothread widget uses EwoksExecutor with execute=None.

payno:
:ok_hand: indeed I miss this one.

src/ewoksorange/gui/qt_utils/app.py
```diff
@@ -179,3 +180,23 @@ def absorb_nonbase_exception(exc_type, exc_value, exc_traceback) -> None:
...
 
     logger.error("Uncaught exception", exc_info=(exc_type, exc_value, exc_traceback))
+
+
+def wait_until(
+    predicate: Callable[[], bool],
+    timeout: float = 5.0,
+    interval: float = 0.01,
+) -> bool:
+    """Wait until ``predicate()`` becomes True while processing Qt events."""
```
payno, line 190:
```suggestion
    """Wait until ``predicate()`` becomes True while processing Qt events.

    :param predicate: Callable returning True when the wait should stop.
    :param timeout: Maximum time to wait, in seconds.
    :param interval: Time to sleep between predicate checks, in seconds.
    :returns: True if the predicate became True before the timeout, False otherwise.
    """
```

src/ewoksorange/tests/executor/signals.py
```diff
@@ -0,0 +1,129 @@
...
+        executor.started.connect(self._store_future("started"))
+        executor.succeeded.connect(self._store_future_and_result("succeeded"))
+        executor.failed.connect(self._store_future_and_result("failed"))
+        executor.aborted.connect(self._store_future("aborted"))
+        executor.finished.connect(self._store_future("finished"))
+        executor.ignored.connect(self._store_noargs("ignored"))
+
+    def _store_future(self, name: str) -> Callable[[TaskFuture], None]:
+        def callback(future: TaskFuture):
+            print(f"Signal {name!r} received")
```
payno, line 43: leftover ?

```suggestion
```

src/ewoksorange/tests/executor/signals.py
```diff
@@ -0,0 +1,129 @@
...
+            self._events[name].append(future)
+            self._order.append(name)
+
+        return callback
+
+    def _store_future_and_result(
+        self, name: str
+    ) -> Callable[[TaskFuture, Union[VariableContainer, Exception]], None]:
+        def callback(future: TaskFuture, result: VariableContainer):
+            print(f"Signal {name!r} received")
```
payno: leftover ?

```suggestion
```

src/ewoksorange/tests/executor/signals.py
```diff
@@ -0,0 +1,129 @@
...
+            self._events[name].append((future, result))
+            self._order.append(name)
+
+        return callback
+
+    def _store_noargs(self, name: str) -> Callable[[], None]:
+        def callback():
+            print(f"Signal {name!r} received")
```
payno, line 51: same ?

```suggestion
```

src/ewoksorange/tests/executor/test_executor_abort.py
```diff
@@ -0,0 +1,29 @@
...
+from ewoksutils.exceptions import TaskExecutionError
+
+from .tasks import AddTask
+
+
+def test_abort(qtapp, executor_context_factory):
+    with executor_context_factory() as (kind, executor, recorder):
+        if kind in ("sync", "thread"):
+            pytest.skip(f"abort not supported by {kind!r} executor")
+            return
```
payno, line 13: Just to be sure: it can be handled but it must be foreseen at the side level right ? (which is not convenient and generic at all I agree)

payno (approved):

src/ewoksorange/tests/executor/signals.py
woutdenolf: I was thinking to keep these prints in case the tests fail. Could be useful to see that the Qt signals were indeed received or not in those cases.

src/ewoksorange/tests/executor/test_executor_abort.py
woutdenolf: I skipped abort because threading does not support abort (threads cannot be killed).

But in fact since currently `abort` does `Task.cancel` it is supported for threads. I'll adapt the test.

woutdenolf: Test that `thread` aborts even-though in the future it might not.

I also fixed a minor detail: `aborted` should return `True` when effectively aborted, not just when abort is requested.

woutdenolf: Also, instead of `pytest.skip` I should check that aborting does NOT work in case of `sync`. Similar in other tests.

woutdenolf: Done. Now when the executor does not support something, instead of `pytest.skip` we effectively check that it does not support the thing.

woutdenolf: The current core API for `cancel` is not clear in fact: 

```python
class Task()

    @property
    def cancelled(self) -> bool:
        """Return True if the task has been cancelled by the user"""
        return self._cancelled

    @cancelled.setter
    def cancelled(self, cancelled: bool) -> None:
        self._cancelled = cancelled

    def cancel(self):
        raise NotImplementedError
```

Does `cancelled == True` mean the cancellation was requested or the task was effectively cancelled?

woutdenolf: Anyway, that's outside the scope of this PR. Currently `TaskFuture.abort()` calls `Task.cancel()` and `TaskFuture.aborted()` returns `Task.cancelled`. Added this to the `TaskFuture` docstrings.

woutdenolf:
Because the way `Task.cancel` is currently implemented (with reset on execution) we have a race condition in the old code (see https://github.com/ewoks-kit/ewoksorange/issues/423) and the new code (this PR).

I will not fix the old code because we will remove it. However I will refactor the controllers and runners in this PR to deal with this.

src/ewoksorange/gui/concurrency/_runners/abstract.py
```diff
@@ -0,0 +1,107 @@
...
+                return
+
+            # Task.execute() resets the task's cancellation state once, right
+            # when execution starts, so a cancel() issued before or during that
+            # reset can be silently undone. Keep re-issuing it until the task
+            # finishes, guaranteeing that at least one call lands after the
+            # reset.
+            while not done.is_set():
+                task.cancel()
+                done.wait(timeout=_RETRY_INTERVAL)
```
woutdenolf: The `Task.cancel()` vs. `Task.reset_state()` race-condition required a major refactor.

Instead of calling `Task.cancel()` upon abort we set the `_abort_event` event.

During the execution of the task we have a thread that waits for `_abort_event` and then calls `Task.cancel()` until we are done.

src/ewoksorange/gui/concurrency/_controllers/_executor.py
```diff
@@ -0,0 +1,68 @@
...
+from typing import Optional
+from typing import Type
+
+from ewokscore.task import Task
+
+from .abstract import TaskController
+
+_logger = logging.getLogger(__name__)
+
+_STARTED_TIMEOUT = 300
```
payno: Could we have the unit somewhere (seconds) ?

src/ewoksorange/gui/concurrency/_runners/abstract.py
payno: I didn't remembered that `reset_state` was existing.

src/ewoksorange/gui/concurrency/_runners/process.py
```diff
@@ -0,0 +1,38 @@
...
+
+_IPC_ERRORS = (EOFError, BrokenPipeError, ConnectionError)
+"""Raised when the manager providing the IPC objects shut down."""
+
+
+class ProcessTaskRunner(ExecutorTaskRunner):
+    """Executes one ewoks task in another process.
+
+    Using the IPC objects is skipped when the manager owning them is gone, so
+    that a task which ran to completion is not reported as failed.
```
payno: If I get it right maybe:

```suggestion
    IPC errors are skipped because raised when the manager owning them is gone, so
    that a task which ran to completion is not reported as failed.
```
else would need a bit more details.

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -0,0 +1,190 @@
...
+    """
+
+    submitted = Signal(TaskFuture)
+    """Emitted when a task is submitted."""
+
+    started = Signal(TaskFuture)
+    """Emitted when a task starts executing."""
+
+    succeeded = Signal(TaskFuture, object)
+    """Emitted when a task finishes successfully. Second argument is the result dict."""
```
payno: The result can be obtained from the `TaskFuture` no ?
Is it useful to emit it as well ?

If we want to emit it a dict should be enough right ?


```suggestion
    succeeded = Signal(TaskFuture, dict)
    """Emitted when a task finishes successfully. Second argument is the result dict."""
```

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -0,0 +1,190 @@
...
+
+    failed = Signal(TaskFuture, object)
+    """Emitted when a task raises an exception. Second argument is the result dict."""
```
payno: Looking at the code this is more

```suggestion
    failed = Signal(TaskFuture, Exception)
    """Emitted when a task raises an exception. Second argument is the exception."""
```
But has for the `succeeded` it can be obtained from the `TaskFuture` so do we need to propagate it also with the signal ?

payno (commented):
Looks good. Need some details on minors API points

src/ewoksorange/gui/concurrency/executor.py
woutdenolf: Yes it is a VariableContainer

src/ewoksorange/gui/concurrency/executor.py
woutdenolf: Indeed, maybe we should just pass the futures.

woutdenolf: Forgot the push the changes. Also had to fix a Task that did not set its `cancelled` state when cancelled. Also fixed a flaky parallelism check logic in the tests.

src/ewoksorange/tests/executor/test_executor_parallel.py
```diff
@@ -0,0 +1,40 @@
...
+        results = [future.result(timeout=10) for future in futures]
+
+        intervals = [(r["start"].value, r["end"].value) for r in results]
+
+        # With 4 tasks and 2 workers, expect two batches of 2 tasks each
+        # running at the same time. The exact pair count is timing-sensitive
+        # (scheduling jitter can make a 3rd task start early), so only the
+        # minimum expected from genuine parallelism is checked.
+        overlapping_pairs = sum(_overlaps(a, b) for a, b in combinations(intervals, 2))
+        assert overlapping_pairs >= workers
```
woutdenolf, line 30: In the previous logic we were testing the overall time of running 4 tasks of 1 second with 2 workers to take less than 3.5 seconds.

This was flaky because it does not take into account the warm-up time of the pool.

In the new test we assume that to 4 time intervals will have at least 2 overlapping pairs. There could be more if task 3 starts before task 2 ends. Nevertheless this confirms that things run in parallel.

PR #422, opened 2026-07-30
owscheme: store ewoks graph attributes into a dedicated new node.

payno:
Move storing ewoks graph attributes from 'description' to a dedicated node.

This is a quick "show case" done by AI. It can be rework if needed. But I don't think we are loosing feature by using a new dedicated node. It will be ignored by orange if we use `orange-canvas` but I think this is the behavior we want.

[demo.ows](https://github.com/user-attachments/files/30533954/demo.xml) result with this PR.
note: demo.ows has been renamed demo.xml to be able to attach the file.

### DIFF
--- a/src/ewoksorange/gui/workflows/owscheme.py
+++ b/src/ewoksorange/gui/workflows/owscheme.py
@@ -12,6 +12,7 @@ from typing import Tuple
 from typing import Type
 from typing import Union
 from uuid import uuid4
+from xml.etree import ElementTree
 
 from ewokscore import load_graph
 from ewokscore.graph import TaskGraph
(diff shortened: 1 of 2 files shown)


src/ewoksorange/gui/workflows/owscheme.py
```diff
@@ -482,6 +484,28 @@ def window_group_presets(self):
         return list()
 
 
+_EWOKS_GRAPH_ATTRS_TAG = "ewoks_graph_attrs"
```
payno, line 495: This would be the new node.

woutdenolf:
When I save the workflow from the GUI, does this special node get preserved?

Edit: just tested and no, it does not. That's the entire reason I was abusing _description_. It needs to be part of a natively supported XML element or attribute.

woutdenolf (changes requested):
Does not fully support graph attributes.

payno:
Ah, indeed, I thought those were already called by orange-canvas.

We can patch the original functions. It's not ideal, but we are already doing it in several places.

Regarding patching the original function versus changing the original goal of the description, I'm leaning 60-40 toward patching. I'll check if it's doable (it should be, from what I see), update the PR, and we can discuss it then.

woutdenolf (approved):
Ok with some more patching. At some point we should look at adding upstream support for some of those.

woutdenolf:
CI is failing though.

woutdenolf:
The additional node would not survive `orange-canvas`. Up to now `orange-canvas` would work for all our workflows. That's a rather major break I would say.

woutdenolf:
Anyway, this is blocking lots of other releases (ewoks, ewoksjob, ewoksserver) so I want a fix that can be deployed as `rc` at least.

woutdenolf:
Failing test is the `Task.cancel` race-condition https://github.com/ewoks-kit/ewoksorange/issues/423 so unrelated.

woutdenolf:
Thanks!

PR #428, opened 2026-08-01
OWEwoksWidgetNoThread uses the new EwoksExecutor

woutdenolf:
`OWEwoksWidgetNoThread` now uses the new `EwoksExecutor`and the old executor code is deleted.

Both `OWEwoksWidgetNoThread` and `_OWEwoksExecutorWidget` now use `EwoksExecutor`.

I might want to refactor those two again after this PR to have one single base that uses `EwoksExecutor`.


woutdenolf:
Failing test unrelated: https://github.com/ewoks-kit/ewoksorange/issues/429

Issue #429, opened 2026-08-01
Flaky test_failed_workfow


woutdenolf:
```
_____________________________ test_failed_workfow ______________________________

sqlite_path = PosixPath('/tmp/pytest-of-runner/pytest-0/test_failed_workfow0')
ewoks_orange_canvas = <ewoksorange.gui.canvas.handler.OrangeCanvasHandler object at 0x7fedb83069d0>

    def test_failed_workfow(sqlite_path, ewoks_orange_canvas):  # noqa 811
        database = sqlite_path / "ewoks_events.db"
>       run_failed_workfow(
            database,
            _execute_graph,
            tempdir=sqlite_path,
            canvas_handler=ewoks_orange_canvas,
        )

/opt/hostedtoolcache/Python/3.9.25/x64/lib/python3.9/site-packages/ewoksorange/tests/test_workflow_events.py:25: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/hostedtoolcache/Python/3.9.25/x64/lib/python3.9/site-packages/ewokscore/tests/test_workflow_events.py:233: in run_failed_workfow
    _execute_graph(database, graph, execute_graph, **execute_options)
/opt/hostedtoolcache/Python/3.9.25/x64/lib/python3.9/site-packages/ewokscore/tests/test_workflow_events.py:306: in _execute_graph
    execute_graph(graph, **execute_options)
/opt/hostedtoolcache/Python/3.9.25/x64/lib/python3.9/site-packages/ewoksorange/tests/test_workflow_events.py:43: in _execute_graph
    canvas_handler.wait_widgets(timeout=10, raise_error=False)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = <ewoksorange.gui.canvas.handler.OrangeCanvasHandler object at 0x7fedb83069d0>
timeout = 10, raise_error = False

ERROR    orangecanvas.registry.discovery:discovery.py:162 An exception occurred while processing EntryPoint(name='Ewoks Test', value='orangecontrib.ewokstest', group='orange.widgets').
Traceback (most recent call last):
[...]

PR #430, opened 2026-08-01
Downstream Orange widgets are cleared when a Orange widget fails

woutdenolf:


src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -571,13 +571,16 @@ def clear_downstream(self) -> None:
...
+        # downstream nodes are always invalidated regardless of the outcome.
         if ORANGE_VERSION == ORANGE_VERSION.oasys_fork:
-            for ewoksname in self.get_task_outputs(exclude_hidden=True):
+            for ewoksname in self.get_output_names(exclude_hidden=True):
                 output = self._get_output_signal(ewoksname)
                 self.send(output.name, invalid_data.INVALIDATION_DATA)
                 # Note: perhaps `self.invalidate(output.name)` is equivalent
         else:
-            for ewoksname in self.get_task_outputs(exclude_hidden=True):
+            for ewoksname in self.get_output_names(exclude_hidden=True):
```
woutdenolf, line 583: Explanation in https://github.com/ewoks-kit/ewoksorange/issues/429#issuecomment-5151306608 why we used to always have task outputs.

I think it is good to not rely on this regardless and use the output names, not the keys in the dynamically set outputs upon task execution.

PR #444, opened 2026-08-02
CI: test scripts/ewoks_executor_demo.py

woutdenolf:
To make sure the manual `EwoksExecutor` demo script does not go out-of-sync with the code.

### DIFF
--- a/scripts/ewoks_executor_demo.py
+++ b/scripts/ewoks_executor_demo.py
@@ -2,6 +2,7 @@ import argparse
 import logging
 import os
 import sys
+import threading
 import time
 from concurrent.futures import ProcessPoolExecutor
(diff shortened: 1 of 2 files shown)


scripts/ewoks_executor_demo.py
```diff
@@ -71,6 +76,9 @@ def __init__(
 
         self.pending = QListWidget()
         self.pending_items: Dict[int, QListWidgetItem] = {}
+        self.buttons: Dict[str, QPushButton] = {}
+
+        self.total_tasks: int = 0
```
payno: A detail but I would go more for `total_tasks_executed` and move to the protected scope (even if this is an example).


```suggestion
        self._total_tasks_executed: int = 0
```

payno (approved):

PR #445, opened 2026-08-03
_OWEwoksExecutorWidget: call progressBarFinished() after propagate_downstream()

### LINKED ISSUE #443: Waiting for the workflow to be over with wait_widgets is flaky
Currently `wait_widgets` is using these to check that everything has been settled

- `signal_manager.has_pending()`
- `signal_manager.is_active(node)`
- `widget.task_executor.is_running()`
- `settled_streak >= 2`

This is [still not enough](https://gitlab.esrf.fr/workflow/ewoksapps/ewoksfluo/-/jobs/2744327) because there is a time between `widget.task_executor.is_running()` and the downstream triggering of the result. So for the workflow look briefly look done but then triggers again.

woutdenolf:
In addition to the fix, I added docs on the "Execution Lifecycle" of an Orange workflow.

The two mermaid diagrams in the docs (LLM used to track the calls and make the diagrams):

1. One execution cycle

image

2. Node “settledness”

image

### DIFF
--- a/src/ewoksorange/gui/owwidgets/threaded.py
+++ b/src/ewoksorange/gui/owwidgets/threaded.py
@@ -91,11 +91,16 @@ class _OWEwoksExecutorWidget(_OWEwoksThreadedBaseWidget, **ow_build_opts):
         self.__last_task_succeeded = True
         self.__last_task_done = True
         self.__last_task_exception = None
-        self.progressBarFinished()
+        # `propagate_downstream` must run before `progressBarFinished`: the
+        # latter flips `signal_manager.is_active(node)` to False, which is
+        # what `wait_widgets`-style polling relies on to know this widget is
+        # done. Clearing it first would let such polling observe "not
+        # active" before the outputs were actually sent downstream.
         try:
             if propagate:
                 self.propagate_downstream(succeeded=True)
         finally:
+            self.progressBarFinished()
             self._output_changed()
 
     def __on_failed(self, task_future: TaskFuture) -> None:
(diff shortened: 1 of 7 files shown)


src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -91,11 +91,16 @@ def __on_succeeded(self, task_future: TaskFuture) -> None:
...
+        # `propagate_downstream` must run before `progressBarFinished`: the
+        # latter flips `signal_manager.is_active(node)` to False, which is
+        # what `wait_widgets`-style polling relies on to know this widget is
+        # done. Clearing it first would let such polling observe "not
+        # active" before the outputs were actually sent downstream.
         try:
             if propagate:
                 self.propagate_downstream(succeeded=True)
         finally:
+            self.progressBarFinished()
```
woutdenolf, line 103: Fix 1

src/ewoksorange/gui/owwidgets/threaded.py
```diff
@@ -104,11 +109,12 @@ def __on_failed(self, task_future: TaskFuture) -> None:
         self.__last_task_succeeded = False
         self.__last_task_done = True
         self.__last_task_exception = task_future.exception()
-        self.progressBarFinished()
+        # See ordering note in `__on_succeeded`.
         try:
             if propagate:
                 self.propagate_downstream(succeeded=False)
         finally:
+            self.progressBarFinished()
```
woutdenolf, line 117: Fix 2

payno (approved):
Thank you very much.
Especially for the diagram. It can help a lot.

PR #450, opened 2026-08-05
use OWEwoksBaseWidget.has_pending_task() instead of EwoksExecutor.is_running()

woutdenolf:
`EwoksExecutor.is_running` was removed and replaced by `OWEwoksBaseWidget.has_pending_task`.

`is_running` cleared on the worker thread the instant the future finished, before its succeeded/failed signal even reached the GUI thread. `has_pending_task()` fixes that for `wait_widgets()`.

The doc under "node settledness" was modified to reflect that the "settledness state" exists in 3 locations and when changing one state, ensure first that the downstream state is changed.

### DIFF
--- a/src/ewoksorange/gui/concurrency/executor.py
+++ b/src/ewoksorange/gui/concurrency/executor.py
@@ -76,17 +76,6 @@ class EwoksExecutor(QObject):
 
         self._manager: Optional[multiprocessing.managers.SyncManager] = None
 
-    @property
-    def is_running(self) -> bool:
-        """Whether a task is currently submitted or executing.
-
-        Unlike the Qt `started`/`succeeded`/`failed` signals, this reflects
-        the submission state synchronously, without waiting for a
-        (possibly cross-thread, queued) signal to be delivered.
-        """
-        with self._lock:
-            return self._running > 0
-
     def submit_task(
         self, task_class: Type[Task], **task_kwargs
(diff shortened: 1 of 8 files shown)


woutdenolf:
FYI the dowstream CI that failed and now succeeds with this PR: https://gitlab.esrf.fr/workflow/ewoksapps/ewoksndreg/-/jobs/2752016

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -76,17 +76,6 @@ def __init__(
...
-    def is_running(self) -> bool:
-        """Whether a task is currently submitted or executing.
-
-        Unlike the Qt `started`/`succeeded`/`failed` signals, this reflects
-        the submission state synchronously, without waiting for a
-        (possibly cross-thread, queued) signal to be delivered.
-        """
-        with self._lock:
-            return self._running > 0
-
```
payno, line 89: I think it was not on the 5.0 version right ?
Else we should consider deprecating it (and falling back to `has_pending_task`)

payno (approved):

src/ewoksorange/gui/concurrency/executor.py
woutdenolf: This was something new so not need for deprecation.

PR #451, opened 2026-08-05
make `DataViewer` HDF5 mode/locking configurable

### LINKED ISSUE #410: make `DataViewer`'s HDF5 open mode/locking configurable
Related `ewoksxas` issue: https://gitlab.esrf.fr/workflow/ewoksapps/ewoksxas/-/work_items/39

## Where

`ewoksorange.gui.widgets.data_viewer.DataViewer`
(`ewoksorange/gui/widgets/data_viewer.py`), method `updateFile`.

Verified against the installed `ewoksorange==5.0.1`:

```python
def updateFile(self, filename):
    if not os.path.exists(filename):
        return
    h5file = self.__getFileObject(filename)
    if h5file is not None:
        self.__refreshAction.trigger()
        return
    self.closeFile(filename)
    model = self.__treeview.findHdf5TreeModel()
    h5file = h5py.File(filename, mode="a")   # <-- hardcoded
    ...
```

`mode="a"` is hardcoded with no way for a project using `DataViewer` to override
it, and no `locking=` argument is passed either, so the default (locking
enabled) always applies.

## Problem

Projects embedding `DataViewer` for read-only browsing (e.g. inspecting an
HDF5 file that may be actively written by another process) have no way to:

1. Open the file read-only instead of append/read-write.
2. Disable HDF5 file locking to avoid contending with another process's lock
   on the same file.

## Proposed change

Add configurability at construction time, with the *current* behaviour kept as
the default so nothing else in `ewoksorange`/downstream breaks:

```python
class DataViewer(qt.QWidget):
    def __init__(self, parent, open_mode: str = "a", locking: bool = True):
        ...
        self.open_mode = open_mode
        self.locking = locking
```

And use it in `updateFile`:

```python
h5file = h5py.File(filename, mode=self.open_mode, locking=self.locking)
```

### Notes for implementation

- Default values (`open_mode="a"`, `locking=True`) preserve current behaviour
  for existing callers.
- Setting `open_mode="r"` together with `locking=False` is the case we need
  downstream (`ewoksxas`), to browse a file without contending with another
  process's write lock.

## Acceptance criteria

- `DataViewer(parent)` behaves exactly as it does today (append mode, locking
  enabled).
- `DataViewer(parent, open_mode="r", locking=False)` opens files read-only
  without taking an HDF5 file lock, and can browse a file already open
  (read or write) in another process.
- `self.open_mode` / `self.locking` are read by `updateFile` (and any other
  call site currently hardcoding `mode="a"`, if more exist) rather than the
  literal being duplicated.

tj-eldridge:
closes #410 

* added `mode` and `locking` kwargs to `DataViewer.__init__`
* added `self._mode` and `self._locking` to `DataViewer`
* added tests that verify mode and locking state

### DIFF
--- a/src/ewoksorange/gui/widgets/data_viewer.py
+++ b/src/ewoksorange/gui/widgets/data_viewer.py
@@ -440,7 +440,11 @@ class DataViewer(qt.QWidget):
         except TypeError:
             # Support silx<2.0.0
             model.sigH5pyObjectLoaded.emit(h5file)
-        model.insertH5pyObject(h5file, filename=filename)
+        try:
+            model.insertH5pyObject(h5file, filename=filename)
+        except TypeError:
+            # Support silx<2.0.0
+            model.insertH5pyObject(h5file)
 
     def setContentSorted(self, sort):
         """Set whether file content should be sorted or not.
--- a/src/ewoksorange/gui/widgets/data_viewer.py
+++ b/src/ewoksorange/gui/widgets/data_viewer.py
@@ -45,7 +45,7 @@ class DataViewer(qt.QWidget):
         viewer.closeAll()
     """
 
-    def __init__(self, parent, *, mode: str = "a", locking: bool | None = None):
+    def __init__(self, parent, *, mode: str = "a", locking: Optional[bool] = None):
         super().__init__(parent)
 
         self._h5files = list()
--- a/src/ewoksorange/gui/widgets/data_viewer.py
+++ b/src/ewoksorange/gui/widgets/data_viewer.py
@@ -45,7 +45,7 @@ class DataViewer(qt.QWidget):
         viewer.closeAll()
     """
 
-    def __init__(self, parent):
+    def __init__(self, parent, *, mode: str = "a", locking: bool | None = None):
         super().__init__(parent)
 
         self._h5files = list()
@@ -78,6 +78,9 @@ class DataViewer(qt.QWidget):
         self.__treeview.activated.connect(self.displaySelectedData)
         self.__treeview.addContextMenuCallback(self.treeContextMenu)
         self.__customizeTreeModelColumns()
+
+        self._mode = mode
+        self._locking = locking
 
     def __setLayout(self, mainWidget):
         layout = qt.QVBoxLayout()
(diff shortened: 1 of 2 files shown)


woutdenolf (approved):

woutdenolf:
LGTM but CI fails on Windows. Please fix. Thanks!

src/ewoksorange/gui/widgets/data_viewer.py
```diff
@@ -45,7 +45,7 @@ class DataViewer(qt.QWidget):
         viewer.closeAll()
     """
 
-    def __init__(self, parent):
+    def __init__(self, parent, *, mode: str = "a", locking: Optional[bool] = None):
```
payno, line 48: Having a class called `DataViewer` with a default mode set to "append" is unexpected.

However, these settings are only used when updateFile is called, which makes more sense.

Including all options in the constructor can quickly make it overloaded.

The `Qt` way of doing things IMO would be more to define default values in the constructor and use specific setters/getters for customization.


```python
    def __init__(self, parent, *, mode: str = "a", locking: Optional[bool] = None):
        ...
        self._updateOpenMode = "a"
        self._updateFileLocking = None

    def setUpdateFileOpenMode(self, mode):
        self._mode = mode

    def setUpdateFileLocking(self, locking):
        self._locking = locking
```

payno (changes requested):
Thanks for the PR :)

We can discuss the comment if not clear. But I think something closer to `Qt` spirit could be done.

src/ewoksorange/gui/widgets/data_viewer.py
woutdenolf: FYI `mode = "a"` was already the hard-coded default.

This has typically been used for widgets looking at results. Since the workflow needs `mode="a"` to save the results and since we cannot open the same file with two different modes, the shortcut was to hard-code `mode="a"`. Now `ewoksxas` will use it to look at the raw data. Hence the PR.

payno: After talking about it with Wout indeed if we add getter / setter we need to handle changing the mode if a file is already open. That we don't want to do. So let's keep it as it is.

payno (approved):

tj-eldridge:
> LGTM but CI fails on Windows. Please fix. Thanks!

The tests now work on Windows.

PR #454, opened 2026-08-06
move progress and EwoksExecutor to OWEwoksBaseWidget which is no longer abstract

### LINKED ISSUE #434: Since EwoksExecutor abstracts the execution strategy: simplify further the OW widget base classes
After the introduction of `EwoksExecutor`, which supports synchronous, thread and process pools, the current class hierarchy looks like this:

```mermaid
flowchart TD
    A[OWEwoksBaseWidget]

    A --> B[OWEwoksWidgetNoThread]
    B -. uses .-> T[EwoksExecutor]

    A --> C[_OWEwoksThreadedBaseWidget]
    C --> D[_OWEwoksExecutorWidget]
    D -. uses .-> T

    D -->|"config only"| E[OWEwoksWidgetOneThread]
    D -->|"config only"| F[OWEwoksWidgetOneThreadPerRun]
    D -->|"config only"| G[OWEwoksWidgetWithTaskStack]
```

Since `EwoksExecutor` now abstracts the execution strategy, I think it should be possible to simplify the hierarchy to something like this:

```mermaid
flowchart TD
    A[OWEwoksBaseWidget]
    A -. uses .-> T[EwoksExecutor]

    A -->|"config only"| B[OWEwoksWidgetNoThread]
    A -->|"config only"| E[OWEwoksWidgetOneThread]
    A -->|"config only"| F[OWEwoksWidgetOneThreadPerRun]
    A -->|"config only"| G[OWEwoksWidgetWithTaskStack]
```

`OWEwoksBaseWidget` could get two class arguments:

```bash
class MyOwWidget(OWEwoksBaseWidget, ewokstaskclass=TaskA, max_workers=..., submit_policy=...):
    ...
```

We can keep the classes `OWEwoksWidgetNoThread`, `OWEwoksWidgetOneThread`, `OWEwoksWidgetOneThreadPerRun` and `OWEwoksWidgetWithTaskStack` if we want but the apps do not need to use them.

woutdenolf:
`OWEwoksBaseWidget` now has everything: task executor and progress handling.

_Task progress_ handling for sub-processes needed another queue `progress_queue` similar to `started_queue`.

FYI _abort_ is not changed in this PR.

### DIFF
--- a/doc/explanations/execution.rst
+++ b/doc/explanations/execution.rst
@@ -65,6 +65,17 @@ One execution cycle
 
         Note over SM: 100ms timer (re-armed), next node's turn
 
+The diagram shows the default case, where the task runs on a background
+thread. The other backends only change who runs the task:
+
+``concurrency="sync"``
+    Every step runs on the GUI thread, inside the ``submit_task()`` call, so
+    ``handleNewSignals()`` only returns once the outputs have been propagated.
+``concurrency="process"``
+    A worker process runs the task. Its lifecycle events and progress travel
+    back over ``multiprocessing`` proxies, relayed onto the GUI thread by
+    :class:`~ewoksorange.gui.concurrency._controllers.process.ProcessTaskController`.
+
 Node "settledness"
 -------------------
 
--- a/doc/tutorials/ewoks_widgets.rst
+++ b/doc/tutorials/ewoks_widgets.rst
@@ -3,7 +3,9 @@
 Ewoks widgets and execution
 ===========================
 
-There are several ways of defining how your Orange Widget will handle the execution of its associated Ewoks task.
+All Ewoks widgets derive from :class:`~ewoksorange.gui.owwidgets.base.OWEwoksBaseWidget`,
+which executes the Ewoks task as :ref:`configured with class arguments
+<design configuration>`. The most common configurations also have a named class:
 
 * :ref:`design qt main thread (OWEwoksWidgetNoThread)`: simple, robust. Long processings can prevent the GUI from responding.
 * :ref:`design single thread no stack (OWEwoksWidgetOneThread)`: execution is separate from the GUI thread. Can only handle one task at once.
@@ -15,18 +17,54 @@ The choice of design depends on your use case: for example, if you deal with sma
 the first design (the simplest one) is the best. Other designs allow more flexibility but are more complex.
 
 
-.. table:: Differences between Ewoks widgets
+.. list-table:: Differences between Ewoks widgets
+   :header-rows: 1
    :widths: auto
 
-   ============================  =======================================  ============================    
-     Widget                       GUI is responsive during execution      Tasks can be run in parallel
-   ============================  =======================================  ============================  
-   OWEwoksWidgetNoThread            No                                     No                       
-   OWEwoksWidgetOneThread           Yes                                    No                       
-   OWEwoksWidgetOneThreadPerRun     Yes                                    Yes                        
-   OWEwoksWidgetWithTaskStack       Yes                                    Yes                        
-   No widget                        Depends on implementation              Depends on implementation 
-   ============================  =======================================  ============================  
+   * - Widget
+     - GUI is responsive during execution
+     - Tasks can be run in parallel
+   * - :ref:`OWEwoksWidgetNoThread <design qt main thread (OWEwoksWidgetNoThread)>`
+     - No
+     - No
+   * - :ref:`OWEwoksWidgetOneThread <design single thread no stack (OWEwoksWidgetOneThread)>`
+     - Yes
+     - No
+   * - :ref:`OWEwoksWidgetOneThreadPerRun <design several thread (OWEwoksWidgetOneThreadPerRun)>`
+     - Yes
+     - Yes
+   * - :ref:`OWEwoksWidgetWithTaskStack <design single thread and stack (OWEwoksWidgetWithTaskStack)>`
+     - Yes
+     - Yes
+   * - :ref:`Native Orange widget <design free implementation (No direct ewoks inheritance)>`
+     - Depends
+     - Depends
+
+.. _design configuration:
+
+Configure the base widget
+-------------------------
+
+Instead of picking one of the classes above, the execution strategy can be given
+directly to :class:`~ewoksorange.gui.owwidgets.base.OWEwoksBaseWidget`
+
+.. code-block:: python
+
+    from ewoksorange.gui.owwidgets.base import OWEwoksBaseWidget
+    from ewokscore.tests.examples.tasks.sumtask import SumTask
+
+    class OWSumTask(
+        OWEwoksBaseWidget,
+        ewokstaskclass=SumTask,
+        concurrency="process",   # "sync", "thread" (default) or "process"
+        max_workers=2,           # None for the pool default
+        submit_policy="always",  # or "drop_if_busy" while the executor is busy
+        mp_context="spawn",      # "spawn" (default), "fork", "forkserver" or None
+    ):
+        pass
+
+Class arguments that are not provided are inherited, so a subclass can change one
+knob without repeating the others.
 
 
 .. _design qt main thread (OWEwoksWidgetNoThread):
@@ -107,9 +145,11 @@ For this, make the Orange widget inherit from :class:`OWEwoksWidgetOneThread`
 
 
 
-The Orange widget is holding a processing thread (`_processingThread`) that will execute the `ewokstaskclass`.
+The Orange widget holds a single-threaded executor
+(:attr:`~ewoksorange.gui.owwidgets.base.OWEwoksBaseWidget.task_executor`) that will
+execute the `ewokstaskclass`.
 
-.. note:: 
+.. note::
     
     The thread can only execute one task at a time: it will refuse to execute further tasks if the current task is still executing. 
     
@@ -176,8 +216,6 @@ To access it you can create a widget inheriting from :class:`OWEwoksWidgetWithTa
 
         want_main_area = False
 
-
-The :class:`SumListWithTaskStack` holds an instance of `progress` in its task arguments.
 
 .. _design free implementation (No direct ewoks inheritance):
 
--- a/src/ewoksorange/gui/concurrency/_progress.py
+++ b/src/ewoksorange/gui/concurrency/_progress.py
@@ -1,0 +1,31 @@
+"""
+Ewoks task progress across a process boundary.
+"""
+
+import queue
+
+from ewokscore.progress import BasePercentageProgress
+
+PROGRESS_STOP = "__stop__"
+"""Sentinel put on the queue by the parent process to stop the relay."""
+
+
+class QueueProgress(BasePercentageProgress):
+    """Picklable progress that forwards updates to the parent process."""
+
+    _TRANSIENT_ERRORS = (EOFError, BrokenPipeError, ConnectionError)
+    # The manager providing the queue shut down.
+
+    def __init__(self, progress_queue: queue.Queue):
+        """
+        :param progress_queue: A `multiprocessing` manager queue proxy.
+        """
+        super().__init__()
+        self._progress_queue = progress_queue
+
+    def _update(self) -> None:
+        """Send the current progress to the parent process."""
+        try:
+            self._progress_queue.put(self._progress)
+        except self._TRANSIENT_ERRORS:
+            pass
--- a/src/ewoksorange/gui/concurrency/executor.py
+++ b/src/ewoksorange/gui/concurrency/executor.py
@@ -2,24 +2,30 @@ from __future__ import annotations
 
 import logging
 import multiprocessing
+import multiprocessing.context
 import multiprocessing.managers
 import weakref
 from concurrent import futures
 from enum import Enum
 from enum import auto
+from queue import Queue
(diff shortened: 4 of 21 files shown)


doc/explanations/execution.rst
```diff
@@ -65,6 +65,17 @@ One execution cycle
...
 
+The diagram shows the default case, where the task runs on a background
+thread. The other backends only change who runs the task:
+
+``concurrency="sync"``
+    Every step runs on the GUI thread, inside the ``submit_task()`` call, so
+    ``handleNewSignals()`` only returns once the outputs have been propagated.
+``concurrency="process"``
+    A worker process runs the task. Its lifecycle events and progress travel
+    back over ``multiprocessing`` proxies, relayed onto the GUI thread by
```
payno: A details but we could imagine using it without the GUI. So maybe the following is more accurate


```suggestion
    Every step runs on the Qt main thread, inside the ``submit_task()`` call, so
    ``handleNewSignals()`` only returns once the outputs have been propagated.
``concurrency="process"``
    A worker process runs the task. Its lifecycle events and progress travel
    back over ``multiprocessing`` proxies, relayed onto the Qt main thread by
```

doc/tutorials/ewoks_widgets.rst
```diff
@@ -15,18 +17,54 @@ The choice of design depends on your use case: for example, if you deal with sma
...
+
+.. code-block:: python
+
+    from ewoksorange.gui.owwidgets.base import OWEwoksBaseWidget
+    from ewokscore.tests.examples.tasks.sumtask import SumTask
+
+    class OWSumTask(
+        OWEwoksBaseWidget,
+        ewokstaskclass=SumTask,
+        concurrency="process",   # "sync", "thread" (default) or "process"
```
payno: Default first
```suggestion
        concurrency="process",   # "thread" (default), "process" or "sync"
```

src/ewoksorange/gui/concurrency/_progress.py
```diff
@@ -0,0 +1,31 @@
...
+        :param progress_queue: A `multiprocessing` manager queue proxy.
+        """
+        super().__init__()
+        self._progress_queue = progress_queue
+
+    def _update(self) -> None:
+        """Send the current progress to the parent process."""
+        try:
+            self._progress_queue.put(self._progress)
+        except self._TRANSIENT_ERRORS:
```
payno, line 30: I guess the `_TRANSIENT_ERRORS` can happen in the case of a multiprocessing Queue correct ?
But I cannot find doc about those different errors (raised by calling `put`). 
Could you elaborate on the code the different use cases that could raise those exception and eventually - if not obvious - why we want to ignore them please.

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -31,12 +38,49 @@ class SubmitPolicy(Enum):
...
+    :param mp_context: Multiprocessing context, only used for `Concurrency.PROCESS`.
+    :return: The executor, or `None` for `Concurrency.SYNC`.
+    """
+    if concurrency is Concurrency.SYNC:
+        return None
+    if concurrency is Concurrency.PROCESS:
+        return futures.ProcessPoolExecutor(
+            max_workers=max_workers, mp_context=mp_context
+        )
+    return futures.ThreadPoolExecutor(max_workers=max_workers)
```
payno: To be honest I don't which other concurrency policy could be added but might be wise to be ready to raise an error the day it comes.


```suggestion
     if concurrency is Concurrency.THREAD:
        return futures.ThreadPoolExecutor(max_workers=max_workers)
    raise ValueError(f"Concurrency policy {concurrency} is not handled")
```

src/ewoksorange/gui/concurrency/executor.py
```diff
@@ -166,9 +233,43 @@ def _submit_process(self, task_class: Type[Task], task_kwargs: dict) -> TaskFutu
...
+        :return: The task arguments to submit.
+        """
+        if "progress" not in task_kwargs:
+            return task_kwargs
+
+        task_kwargs = dict(task_kwargs)
+        progress = task_kwargs.pop("progress")
+
+        if progress is None or not issubclass(task_class, TaskWithProgress):
+            return task_kwargs
```
payno: Several points:
according to typing `task_kwargs` should already be a dict. So I don't think we need `task_kwargs = dict(task_kwargs)`
And I think by calling `pop` first and providing a default value we can simplify a bit:


```suggestion
        progress = task_kwargs.pop("progress", None)

        if progress is None or not issubclass(task_class, TaskWithProgress):
            return task_kwargs
```

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -64,10 +73,60 @@ class OWEwoksBaseWidget(OWWidget, metaclass=OWEwoksWidgetMetaClass, **ow_build_o
...
 
-    - methods: _get_task_outputs, _execute_ewoks_task, has_pending_task
-    - properties: task_succeeded, task_done, task_exception
+    - `Concurrency.SYNC` (`"sync"`): execute in the calling (GUI) thread.
+    - `Concurrency.THREAD` (`"thread"`): execute in a thread pool.
+    - `Concurrency.PROCESS` (`"process"`): execute in a process pool. This requires
+      all task inputs and outputs to be picklable.
+    """
+
+    _MAX_WORKERS: Optional[int] = 1
```
payno: I had in mind that we switch to None for the default.

```suggestion
    _MAX_WORKERS: Optional[int] = None
```

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -84,6 +143,65 @@ def __init__(self, *args, **kwargs):
         ]
         self.__post_task_exception: Optional[Exception] = None
 
+        self.__taskProgress = QProgress()
+        self.__taskProgress.sigProgressChanged.connect(self._onProgressChanged)
```
payno: Shouldn't the creation of the task progress depends on the ewoks Task as well ? (Only if is a `TaskWithProgress`)

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -84,6 +143,65 @@ def __init__(self, *args, **kwargs):
...
+        self.__propagate_by_future: Dict[TaskFuture, bool] = {}
+        self.__propagate_next: bool = False
+
+        self.__last_output_variables: Optional[VariableContainer] = None
+        self.__last_task_succeeded: Optional[bool] = None
+        self.__last_task_done: Optional[bool] = None
+        self.__last_task_exception: Optional[Exception] = None
+
+        # Note: this might be removed in the future. Please avoid using it.
+        self.__current_task_future: Optional[TaskFuture] = None
```
payno: This is a bit weird to have the connection of signals / slots (processing) before definition of the class variables. Could we reorder it ?


```suggestion
        self.__taskProgress = QProgress()

        self.__executor = EwoksExecutor(
            self._create_pool_executor(),
            self._SUBMIT_POLICY,
            mp_context=self._MP_CONTEXT,
        )

        self.__propagate_by_future: Dict[TaskFuture, bool] = {}
        self.__propagate_next: bool = False

        self.__last_output_variables: Optional[VariableContainer] = None
        self.__last_task_succeeded: Optional[bool] = None
        self.__last_task_done: Optional[bool] = None
        self.__last_task_exception: Optional[Exception] = None

        # Note: this might be removed in the future. Please avoid using it.
        self.__current_task_future: Optional[TaskFuture] = None
        
        # Connect signal / slots
        self.__taskProgress.sigProgressChanged.connect(self._onProgressChanged)

        self.__executor.submitted.connect(self.__on_submitted)
        self.__executor.started.connect(self.__on_started)
        self.__executor.succeeded.connect(self.__on_succeeded)
        self.__executor.failed.connect(self.__on_failed)

```

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -634,50 +769,76 @@ def execute_ewoks_task_without_propagation(self) -> Optional[TaskFuture]:
...
+    def task_executor(self) -> EwoksExecutor:
+        """
+        The executor that runs the Ewoks tasks.
+
+        :return: The :class:`EwoksExecutor` instance.
+        """
+        return self.__executor
+
+    @property
     def task_succeeded(self) -> Optional[bool]:
```
payno, line 796: I would also deprecate this function.
Having a function named `task_succeeded` in the context of concurrency doesn't make much sense.

Same for `task_done` and `task_exception`

src/ewoksorange/gui/owwidgets/base.py
```diff
@@ -753,15 +907,86 @@ def __post_task_execute(self, callbacks: List[Callable[[], None]]) -> None:
...
+        self.__last_task_succeeded = True
+        self.__last_task_done = True
+        self.__last_task_exception = None
+        # `propagate_downstream` must run before `progressBarFinished`: the
+        # latter flips `signal_manager.is_active(node)` to False, which is
+        # what `wait_widgets`-style polling relies on to know this widget is
+        # done. Clearing it first would let such polling observe "not
+        # active" before the outputs were actually sent downstream.
+        # `has_pending_task()` must stay True until both of those have run,
+        # for the same reason, so the future is only popped last.
```
payno, line 1010: 👍

src/ewoksorange/gui/owwidgets/meta.py
```diff
@@ -21,29 +25,148 @@
...
+    ):
+        return mp_context
+    available = multiprocessing.get_all_start_methods()
+    if not isinstance(mp_context, str) or mp_context not in available:
+        names = ", ".join(repr(name) for name in available)
+        raise ValueError(
+            f"'mp_context' must be None, a multiprocessing context or one of {names} on this platform: {mp_context!r}"
+        )
+    return multiprocessing.get_context(mp_context)
+
```
payno, line 167: Maybe this could be part of some utils module but no strong opinion on it.
Could you ask claude to create some test for it ?

src/ewoksorange/gui/concurrency/_progress.py
woutdenolf:
```
_TRANSIENT_ERRORS = (EOFError, BrokenPipeError, ConnectionError)
# The manager providing the queue shut down.
```

Also exists in `ProcessTaskRunner` and I found this issue when shutting down the test app

```
python scripts/ewoks_executor_demo.py
```

This script reproduces `EOFError` and `BrokenPipeError`

```python
import multiprocessing as mp
from queue import Empty
import time


def worker(queue):

    while True:
        try:
            print(queue.get(timeout=0.1))
        except Empty:
            pass
        except Exception as e:
            print(
                f"Worker: {type(e).__name__}: {e!r}",
                flush=True,
            )
            time.sleep(0.1)


def main():
    manager = mp.Manager()
    queue = manager.Queue()

    worker_process = mp.Process(
        target=worker,
        args=(queue,),
    )
    worker_process.start()

    time.sleep(1)

    queue.put("hello")

    print("Main: shutting down the manager!")
    manager.shutdown()
    del manager

    print("Main: manager is gone")

    time.sleep(1)

    worker_process.terminate()
    worker_process.join()


if __name__ == "__main__":
    main()
```

```
Main: shutting down the manager!
hello
Worker: EOFError: EOFError()
Main: manager is gone
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
Worker: BrokenPipeError: BrokenPipeError(32, 'Broken pipe')
```

`ConnectionError` is the base error of `BrokenPipeError` which I added to be sure

```
OSError
  └── ConnectionError
       ├── BrokenPipeError
       ├── ConnectionAbortedError
       ├── ConnectionRefusedError
       └── ConnectionResetError
```

src/ewoksorange/gui/owwidgets/meta.py
woutdenolf: Those are only used in this module so I would keep them there.

I will make tests.

I noticed I forgot to remove `ABCMeta`.

src/ewoksorange/gui/owwidgets/base.py
woutdenolf: Done

src/ewoksorange/gui/owwidgets/meta.py
woutdenolf: Done.

woutdenolf:
More CI trouble: https://github.com/ewoks-kit/ewoksorange/actions/runs/31482007622/job/93748946084?pr=454

Edit: fix in EwoksExecutor -> lazy _manager init guarded.

src/ewoksorange/gui/owwidgets/base.py
woutdenolf: Done

PR #455, opened 2026-08-07
fix[orange link]: #447 Valid link should match pydantic logic in Oran…

### LINKED ISSUE #447: Inconsistency between Pydantic model and Orange
My issue in Darfix:

I have an output parameter `a` of class `A`.

I have an input parameter `b` of class `B(A)`.

With orange current version i can connect `a` output to `b` input.

First i would expect this is not working.

Second there is an inconsistency with Pydantic and my task fail when I run it as pydantic does not accept an input of class `A` (And this is what i'm expecting).

Proposition of a fix:

`src/ewoksorange/gui/orange_utils/_signals.py`:

```
            if is_input:
                signal = signal_class(name=orangename, type=data_type, doc=doc)
            else:
                # To match with Pydantic Output and Input types must match exactly
                # dynamic by default is True in Orange. It means Orange accept a connection if Output type is a base type
                # of input type.
                # We do not want that so we add `dynamic=False`
                signal = signal_class(name=orangename, type=data_type, doc=doc, dynamic=False)


```

@payno @loichuder @woutdenolf ? What do you think ?

maxenceprog:


loichuder:
For the record, I had a branch to test things out in https://github.com/ewoks-kit/ewoksorange/tree/use-non-dynamic-links. Feel free to use it or not.

woutdenolf:
Sorry I pushed a commit to the wrong branch. I removed it again.

woutdenolf:
Is this ready for review?

woutdenolf:
For reference:

We always check `issubclass(source_t, sink_types)`

https://github.com/biolab/orange-canvas-core/blob/d24cbdb1245c3083476cd3bff83b8160f3b590bf/orangecanvas/scheme/link.py#L105

When `dynamic=True` we also check `issubclass(sink_t, source_types)`

https://github.com/biolab/orange-canvas-core/blob/d24cbdb1245c3083476cd3bff83b8160f3b590bf/orangecanvas/scheme/link.py#L109

src/ewoksorange/gui/orange_utils/_signals.py
```diff
@@ -286,7 +286,18 @@ def validate_signals(
                 ewoksname,
             )
             orangename = ewoksname
-            signal = signal_class(name=orangename, type=data_type, doc=doc)
+
+            if is_input:
```
woutdenolf:
```suggestion
            if is_input or ORANGE_VERSION == ORANGE_VERSION.oasys_fork:
```

src/ewoksorange/gui/orange_utils/_signals.py
```diff
@@ -286,7 +286,18 @@ def validate_signals(
...
+                signal = signal_class(name=orangename, type=data_type, doc=doc)
+            else:
+                # To match with Pydantic Output and Input types must match exactly
+                # dynamic by default is True in Orange. It means Orange accept a connection if Output type is a base type
+                # of input type.
+                # We do not want that so we add `dynamic=False`
+                # https://github.com/ewoks-kit/ewoksorange/issues/447
```
woutdenolf:
```suggestion
                # For Output and Input type to be considered compatible by Orange:
                # dynamic=False: Output type may be the same as or a subclass of the Input type.
                # dynamic=True: Output type may be the same as, a subclass of, or a superclass of the Input type.
```

maxenceprog:
oups was in holliday last week and completly forgot this PR thanks for the review

maxenceprog:
one test is now broken: 

`src/ewoksorange/tests/workflow/test_dynamic_link.py`

```
            {
                "id": "1",
                "task_generator": "ewoksorange.bindings.taskwrapper.owwidget_task_wrapper",
                "task_identifier": qualname(NativeWidget),
                "task_type": "generated",
            },
```

When input is generated like this, there is no pydantic model so the input type is `object`

In the old code as `object` is a mother class of everything this is successful with the weird "dynamic" orange mechanism.


I'm not sure if I need to reconsider my code or modify the unit test, need to discuss with @payno  as he wrotes the test

payno:
Indeed having `object` is the `default` behavior (no pydantic models). We don't want to force user to have a pydantic model. So I would say we might want to force `dynamic=False` only if pydantic models are there.

woutdenolf:
Ok so now it is clear why Orange does this by default. So as @payno suggested I suppose this should work

```python
dynamic=not issubclass(data_type, BaseModel)
```

woutdenolf (approved):

